PR #5, opened 2018-10-03
travis: push hugo website to gh-pages

greut:
(we need the token in the travis builds anyway)


greut:
@pyr Jenkins builds expects #6 to be merged in.

greut:
(rebased)

PR #22, opened 2018-10-12
Account flag overwritten in exo api

pierre-emmanuelJ:
issue #21 
-A, --account flag is overwritten in some exo api cmd.
help wanted to find another solution

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -110,7 +110,7 @@ func Execute(version, commit string) {
 
 func init() {
 	RootCmd.PersistentFlags().StringVarP(&gConfigFilePath, "config", "C", "", "Specify an alternate config file [env EXOSCALE_CONFIG]")
-	RootCmd.PersistentFlags().StringVarP(&gAccountName, "account", "A", "", "Account to use in config file [env EXOSCALE_ACCOUNT]")
+	RootCmd.PersistentFlags().StringVarP(&gAccountName, "accountt", "A", "", "Account to use in config file [env EXOSCALE_ACCOUNT]")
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -110,7 +110,7 @@ func Execute(version, commit string) {
 
 func init() {
 	RootCmd.PersistentFlags().StringVarP(&gConfigFilePath, "config", "C", "", "Specify an alternate config file [env EXOSCALE_CONFIG]")
-	RootCmd.PersistentFlags().StringVarP(&gAccountName, "account", "A", "", "Account to use in config file [env EXOSCALE_ACCOUNT]")
+	RootCmd.PersistentFlags().StringVarP(&gAccountName, "accountt", "A", "", "Account to use in config file [env EXOSCALE_ACCOUNT]")
```
pierre-emmanuelJ: WIP need some proposition

greut: we remove it and over an easy way to switch between them.

pierre-emmanuelJ: --ACCOUNT can't be fine ?
```suggestion
	RootCmd.PersistentFlags().StringVarP(&gAccountName, "ACCOUNT", "A", "", "Account to use in config file [env EXOSCALE_ACCOUNT]")
```

greut: or longer? `--use-account` ?

pierre-emmanuelJ: yes can be great !
```suggestion
	RootCmd.PersistentFlags().StringVarP(&gAccountName, "use-account", "A", "", "Account to use in config file [env EXOSCALE_ACCOUNT]")
```

PR #23, opened 2018-10-12
Add interactive account selection

pierre-emmanuelJ:
Add account selection:
Able to select one
image

then quit or add another account

image

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -27,13 +27,32 @@ const (
 // configCmd represents the config command
 var configCmd = &cobra.Command{
 	Use:   "config",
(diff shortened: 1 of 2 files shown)


greut:
I'd keep current `config set` as is and improve the `exo config` instead.

```
% exo config
? Select an account
 - exoscale-1 [Default]
   test
   [+] Do you wish to add another account?
```

pierre-emmanuelJ:
No, Look the update of the description. before you could see the account  and add another one
but now you can select and add another one

greut:
@pierre-emmanuelJ my bad, I hit ^C too early.

PR #24, opened 2018-10-12
Add interactive zone selection

pierre-emmanuelJ:
new way to select zone in config:

`exo config`
image

image


cmd/config.go
```diff
@@ -526,57 +527,38 @@ func getAccountByName(name string) *account {
...
+		zones[i] = zName
+	}
 
-		n := fmt.Sprintf("%d", i+1)
-
-		zones[n] = zName
-
-		fmt.Printf("%d: %s\n", i+1, zName)
+	prompt := promptui.Select{
+		Label: fmt.Sprintf("Choose %q default zone", accountName),
```
chrisglass: How about "Choose the default zone for %q" instead?

pierre-emmanuelJ: 👍

PR #34, opened 2018-10-26
Add sos download progress bar

pierre-emmanuelJ:
```
$ exo sos download eip config.go testtest.go
config.go  100 % [==============================================================] 0s
Successfully downloaded config.go into "testtest.go"
```

### DIFF
--- a/cmd/sos_download.go
+++ b/cmd/sos_download.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"io"
 	"log"
-	"math/rand"
 	"os"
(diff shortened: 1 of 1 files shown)


greut:
:fire: :fireworks: !

pierre-emmanuelJ:
for 
`exo sos download mybucket mysubfolder/myfile.txt myfile.txt`
you can do that:
`exo sos download mybucket mysubfolder/myfile.txt .`

but for the path I agree with you :+1:

PR #36, opened 2018-10-29
Build/mac

retrack:
An attempt at creating a [Homebrew tap](https://github.com/exoscale/homebrew-tap
) on build with Goreleaser.

After publication, install on MacOS with

`brew tap exoscale/tap`
`brew install exo`

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -40,3 +40,16 @@ nfpm:
   formats:
     - deb
     - rpm
+
+brew:
+  name: "exo"
+  github:
+    owner: exoscale
+    name: homebrew-exo
+
+  folder: Formula
+  homepage: "https://exoscale.github.io/cli/"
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -40,3 +40,16 @@ nfpm:
...
+  name: "exo"
+  github:
+    owner: exoscale
+    name: homebrew-tap
+  folder: Formula
+  homepage: "https://exoscale.github.io/cli/"
+  description: Manage (almost) everything at Exoscale from the command line.
+  dependencies:
+      - go
+  skip_upload: true
```
greut: - `exo` itself has no dependencies, we can skip this section
- we can also let it upload the Formula as it'll only be done when a release is made

greut:
```console
% brew tap exoscale/tap
% brew install exo
```

PR #49, opened 2018-11-07
status: shorter display

pyr:
We should push people towards the status website and twitter account as early as possible.

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -47,7 +47,7 @@ var statusCmd = &cobra.Command{
 				suffix,
 				status.Incidents[0].Title);
 			fmt.Println(msg)
-			fmt.Printf("Updates available at %s\n", twitterUrl)
+			fmt.Printf("Updates available at %s\n", twitterURL)
 			return fmt.Errorf(msg)
 		}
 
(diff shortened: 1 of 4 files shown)


cmd/status.go
```diff
@@ -38,17 +38,15 @@ var statusCmd = &cobra.Command{
...
-				fmt.Fprintf(w, "\t%s\t%s\n\n", event.Updated, event.Created)             // nolint: errcheck
+			suffix := ""
+			if len(status.Incidents) > 1 {
+				suffix = "s"
 			}
-
-			fmt.Fprintln(w) // nolint: errcheck
-
-			w.Flush()
-			return fmt.Errorf("%d incidents, %s", len(status.Incidents), twitterURL)
```
greut: imho, failing in case of an incident opens some automation possibilities.

cmd/status.go
```diff
@@ -38,17 +38,15 @@ var statusCmd = &cobra.Command{
...
+			fmt.Printf("%d ongoing Incident%s (last: %s)\n",
+				len(status.Incidents),
+				suffix,
+				status.Incidents[0].Title);
+			fmt.Printf("Full updates available at %s\n", twitterURL)
 		}
 
 		w.Flush()
```
greut: the _flushing_ should probably be done earlier then.

PR #53, opened 2018-11-08
Add "lab kube" commands

falzm:
Here is a first version of the `exo kube` command, allowing a user to bootstrap quickly and easily a simple single-node Kubernetes cluster on top of Exoscale.

For now the implementation is minimal: 

* `exo kube create [options] <cluster name>` creates a VM instance, bootstraps the k8s setup inside then retrieves and stores k8s master API endpoint `kubeconfig` locally
* `exo kube env <cluster name>` populates local environment variables to use the `kubeconfig` retrieved earlier
* `exo kube delete <cluster name>` deletes the VM instance hosting the k8s cluster

Let me know what you think about it 🤗


falzm:
TODO:

* upgrade/reboot the system before bootstrapping k8s
* enable master [to host pods](https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/#master-isolation)
* expose remote Docker host locally via `exo kube env` (à la `minikube docker-env`)

greut:
for the record:
- it looks promising! :100: :tada: 
- when ssh-ing into the machine many packages are asking for an upgrade (nit pick)
- running the dashboard fails as the master node is tainted to not run pods
- having something like `minikube docker-env` would enable building and running custom images without having to go through setting up a registry (or using docker hub)

cmd/kube_create.go
```diff
@@ -0,0 +1,312 @@
...
+						Description:     "Kubernetes API",
+						CIDRList:        []egoscale.CIDR{*egoscale.MustParseCIDR("0.0.0.0/0")},
+						Protocol:        "TCP",
+						StartPort:       6443,
+						EndPort:         6443,
+					},
+				}
+
+				for _, rule := range sgRules {
+					if _, err = cs.RequestWithContext(context.TODO(), rule); err != nil {
```
pierre-emmanuelJ: You can use `gContext`

pierre-emmanuelJ: and can be used to make all request with ctx

falzm: Done.

cmd/kube_create.go
```diff
@@ -30,43 +30,69 @@ type kubeBootstrapStep struct {
...
 sudo add-apt-repository \
    "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
    $(lsb_release -cs) \
    stable"
-
 sudo apt-get update && sudo apt-get install -y docker-ce=18.06.0~ce~3-0~ubuntu
 
+curl -so cfssl https://pkg.cfssl.org/R1.2/cfssl_linux-amd64
+curl -so cfssljson https://pkg.cfssl.org/R1.2/cfssljson_linux-amd64
+chmod +x cfssl* && sudo mv cfssl* /usr/bin
```
greut: https://packages.ubuntu.com/bionic/amd64/golang-cfssl/filelist

cmd/kube_create.go
```diff
@@ -30,43 +30,69 @@ type kubeBootstrapStep struct {
...
+exit`},
+	{name: "Docker Engine installation", command: `\
 curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
-
 sudo add-apt-repository \
    "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
    $(lsb_release -cs) \
    stable"
-
 sudo apt-get update && sudo apt-get install -y docker-ce=18.06.0~ce~3-0~ubuntu
```
greut: also https://packages.ubuntu.com/cosmic/docker.io

greut: at the end of the day, it's harder to maintain on our side to do so.

falzm: We're not exactly talking about 1000+ or even 100+ LoC to maintain here... Let's agree to disagree 😏

falzm:
As discussed yesterday, it's been decided to relocate the `kube *` commands under a new `lab` command in order to avoid confusion regarding the stability/durability of this feature.

cmd/kube_env.go
```diff
@@ -0,0 +1,44 @@
...
+			return err
+		}
+
+		vm, err := getVMWithNameOrID(string(clusterInstance))
+		if err != nil {
+			return err
+		}
+
+		fmt.Printf("export KUBECONFIG=\"%s/%s\"\n", getKubeconfigPath(clusterName), "kubeconfig")
+		fmt.Printf("export DOCKER_HOST=\"tcp://%s:2376\"\n", vm.IP().String())
```
greut:
```suggestion
		fmt.Printf(`export DOCKER_HOST="tcp://%s:2376"\n`, vm.IP().String())
```

cmd/kube_list.go
```diff
@@ -0,0 +1,48 @@
...
+	Use:     "list",
+	Short:   "List running Kubernetes cluster instances",
+	Aliases: gListAlias,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return listKubeInstances()
+	},
+}
+
+func listKubeInstances() error {
+	vms, err := cs.ListWithContext(gContext, &egoscale.VirtualMachine{})
```
greut: with egoscale 0.13, you'll have more options here.

```go
vms, err := cs.ListWithContext(gContext, &egoscale.ListVirtualMachines{
    Tag: []egoscale.ResourceTag{{
        Key: "managedby",
        Value: "exokube",
    }}
})
```

greut: in 0.12, you'll have to enrich `VirtualMachine.ListRequest()` response.

cmd/kube_env.go
greut: what about the other ones?

falzm:
How are we doing so far, regarding the feature set?

greut:
`getKubeInstanceVersion` should rely how tags are working in CloudStack, so fixed value + fixed version.

https://github.com/exoscale/cli/pull/53#discussion_r236586324

`GetVMWithNameOrID` is bad as it doesn't properly handle the 41, hopefully egoscale 0.13 will help.

cmd/kube_list.go
falzm: OK, I've put a `FIXME` to update the code once we reach egoscale 0.13.

cmd/kube_create.go
```diff
@@ -0,0 +1,457 @@
...
+// kubeBootstrapSteps represents a k8s instance bootstrap steps
+var kubeBootstrapSteps = []kubeBootstrapStep{
+	{name: "Instance system upgrade", command: `\
+sudo -E DEBIAN_FRONTEND=noninteractive apt-get update && sudo -E DEBIAN_FRONTEND=noninteractive apt-get upgrade -y
+sudo -E DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    apt-transport-https \
+    ca-certificates \
+	curl \
+	golang-cfssl \
+	software-properties-common
```
greut: nitpick; mix of spaces and tabs.

cmd/kube_create.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+}
+EOF
+
+sudo mv ca.pem /etc/docker/ca.pem
+sudo mv cert.pem /etc/docker/cert.pem
+sudo mv cert-key.pem /etc/docker/key.pem
+rm -f *.{csr,json,pem}
+
+sudo sed -i -re 's#^ExecStart=.*#ExecStart=/usr/bin/dockerd#' /lib/systemd/system/docker.service
```
greut: It'd be nicer to use the override mechanism rather than modifying stuff in `/lib/systemd` :cowboy_hat_face: 

https://wiki.archlinux.org/index.php/Docker#Remote_API_with_systemd

cmd/kube_create.go
```diff
@@ -0,0 +1,457 @@
...
+cat <<EOF | sudo tee /etc/apt/sources.list.d/kubernetes.list
+deb http://apt.kubernetes.io/ kubernetes-xenial main
+EOF
+
+sudo -E DEBIAN_FRONTEND=noninteractive apt-get update && \
+sudo -E DEBIAN_FRONTEND=noninteractive apt-get install -y kubelet kubeadm kubectl && \
+sudo apt-mark hold kubelet kubeadm kubectl`},
+	{name: "Kubernetes cluster node initialization", command: `\
+sudo kubeadm init --pod-network-cidr=192.168.0.0/16 --kubernetes-version {{ .Version }} &&
+sudo kubectl --kubeconfig=/etc/kubernetes/admin.conf taint nodes --all node-role.kubernetes.io/master- &&
```
greut: `&&` ?

cmd/kube_create.go
```diff
@@ -0,0 +1,457 @@
...
+		}
+	}
+
+	return ioutil.WriteFile(dst, buf.Bytes(), 0600)
+}
+
+func init() {
+	kubeCreateCmd.PersistentFlags().BoolVarP(&kubeCreateDebug, "debug", "d", false, "debug mode on")
+	kubeCreateCmd.Flags().StringP("version", "v", "stable-1", "<version label> "+
+		"(see https://godoc.org/github.com/kubernetes/kubernetes/cmd/kubeadm/app/util#KubernetesReleaseVersion)")
```
greut: I'd prefer if we had a the real underlying Kubernetes version, e.g. 1.13.0

cmd/kube_create.go
falzm: Fixed.

cmd/kube_create.go
falzm: IMO it's not necessary to setup kube configuration for the `ubuntu` user, we're not supposed to use it.

falzm:
It has been suggested to add a new `exo lab kube ssh <cluster instance>` command in order to facilitate the connection to a running instance. Currently it is possible to do so, but not really user-friendly (i.e. `ssh -i ~/.exoscale/instances/<instance UUID>/id_rsa ubuntu@<instance IP address>`).

greut:
@falzm what about `exo ssh <cluster_instance>` ?

falzm:
> @falzm what about `exo ssh <cluster_instance>` ?

I guess it's better not to leak `exo lab kube` implementation outside of its domain/scope, but in practice the `exo lab kube ssh` will use the `exo vm ssh` command code.

greut:
what's so special about `<cluster instance>` it requires its own set of commands?

greut:
If I understand it correctly, `exo ssh <cluster instance>` would work out of the box.

falzm:
> If I understand it correctly, `exo ssh <cluster instance>` would work out of the box.

Almost:

```console
exo ssh aperikube
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0755 for '/Users/marc/.exoscale/instances/a35b8366-83a7-48b2-979c-5136b9810759' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
Load key "/Users/marc/.exoscale/instances/a35b8366-83a7-48b2-979c-5136b9810759": bad permissions
ubuntu@159.100.242.251's password:
```

cmd/kube_create.go
greut:
```console
% exo lab kube create <foo> -v 1.12.3
```

greut:
When the creation fails, it refuses to be deleted.

```console
% exo lab kube create yoyo -v 1.12.3  
...
        [ERROR KubeletVersion]: the kubelet version is higher than the control plane version. This is not a supported version skew and may lea
d to a malfunctional cluster. Kubelet version: "1.13.1" Control plane version: "1.12.3"                                                       
[preflight] If you know what you are doing, you can make a check non-fatal with `--ignore-preflight-errors=...

% exo lab kube list
┼──────┼─────────────────┼───────┼─────────┼─────────┼
│ NAME │   IP ADDRESS    │ SIZE  │ VERSION │  STATE  │
┼──────┼─────────────────┼───────┼─────────┼─────────┼
│ yoyo │ 159.100.241.202 │ Small │ 1.12.3  │ Running │
┼──────┼─────────────────┼───────┼─────────┼─────────┼

% exo lab kube delete yoyo                                                                                                  
"yoyo": no such cluster

% exo vm delete yoyo
Destroying "yoyo"  100 % [======================================]
```

falzm:
> > If I understand it correctly, `exo ssh <cluster instance>` would work out of the box.
> 
> Almost:

This was actually a bug, fixed in 6cc1d26ff30b9cecafcaf506d46e51d5e144510d – it now works OK with `exo ssh` 👍

nfrankel:
What is the equivalent to `minikube ip`?

falzm:
> What is the equivalent to `minikube ip`?

@nfrankel you can have the master node's IP address by listing the instances:

```console
% exo lab kube list
┼───────────┼─────────────────┼───────┼──────────┼─────────┼
│   NAME    │   IP ADDRESS    │ SIZE  │ VERSION  │  STATE  │
┼───────────┼─────────────────┼───────┼──────────┼─────────┼
│ aperikube │ 159.100.242.251 │ Small │ stable-1 │ Running │
┼───────────┼─────────────────┼───────┼──────────┼─────────┼
```

If it's not convenient enough, I could add a `exo lab kube ip <instance>` subcommand too.

nfrankel:
+1 for `exo lab kube ip <instance>`. Also, it would be great to allow to provide no `<instance>` argument, and to return the IP of the only running instance in that case (or send an error otherwise).

cmd/kube_create.go
```diff
@@ -102,12 +106,20 @@ curl -fsSL https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key
...
-sudo -E DEBIAN_FRONTEND=noninteractive apt-get install -y kubelet kubeadm kubectl && \
+export PKG_VERSION=$(apt-cache madison kubelet | awk '$3 ~ /{{ .Version }}/ { print $3 }' | head -n 1)
+if [[ -z "${PKG_VERSION}" ]]; then
+	echo "error: unable to find package for version {{ .Version }}" >&2
+	exit 1
+fi
+
+sudo -E DEBIAN_FRONTEND=noninteractive apt-get install -y kubelet=${PKG_VERSION} \
+	kubeadm=${PKG_VERSION} \
+	kubectl=${PKG_VERSION} && \
```
greut: what's the risk that the `PKG_VERSION` of `kubelet` isn't the same as the one of `kubeadm` or `kubectl`?

falzm: With this code I'd say `PKG_VERSION` will be the same for all `kubelet`, `kubeadm` and `kubectl`. There may be a problem with the `kubectl` version that the user is supposed to install locally once the cluster instance is bootstrapped, though.

falzm:
> +1 for `exo lab kube ip <instance>`.

Here you go @nfrankel:

```console
% exo lab kube ip 1ptikube
185.19.28.235
```

> Also, it would be great to allow to provide no `<instance>` argument, and to return the IP of the only running instance in that case (or send an error otherwise).

I'm not very fond of this special case, to me it requires quite a lot of code for just a few spared key strokes... 😒  This is one of the few differences with Minikube that I'm willing to stand for.

greut:
Agreed, it probably needs some kind of `exo ip <vm>` command

falzm:
Well, so long wow: https://github.com/gernest/wow/issues/14

greut:
https://sos-ch-dk-2.exo.io/yoanyoan/exoscale-cli_SNAPSHOT-888783455f900823148b72f137ce199a1d0ba566_linux_amd64.tar.gz

falzm:
A last-minute addition suggested by @nfrankel worth considering: install the [Kubernetes Dashboard](https://github.com/kubernetes/dashboard) during instance bootstrap, and provide a `exo lab kube dashboard` subcommand that will start the `kubectl proxy` to the cluster then output the access token to the user, and open a link to the proxy URL directly in the user's browser. WDYT?

greut:
@falzm @nfrankel open an issue. It could be done at later point. This PR is quite old already

falzm:
@greut Ready for merging.

PR #57, opened 2018-11-14
config: allow secret to come from an external command

greut:
Fixes #18 

```toml
[[accounts]]
  key = "EXOc94bc655ff901b7218a5c3cd"
  secretCommand = ["secret-tool", "lookup", "Exoscale", "EXOc94bc655ff901b7218a5c3cd"]
```

```console
% exo config show
Default template: Linux Ubuntu 18.04 LTS 64-bit
Name:         yoan.blanc@exoscale.ch
API Key:      EXOc94bc655ff901b7218a5c3cd
API Secret:   secret-tool lookup Exoscale EXOc94bc655ff901b7218a5c3cd
Account:      yoan.blanc@exoscale.ch
Default zone: ch-gva-2

% exo zone list
```
|   NAME   |                  ID                  |
|----------|--------------------------------------|
| ch-gva-2 | 1747ef5e-5451-41fd-9f1a-58913bae9702 |
| ch-dk-2  | 381d0a95-ed4a-4ad9-b41c-b97073c1a433 |
| at-vie-1 | b0fcd72f-47ad-4779-a64f-fe4de007ec72 |
| de-fra-1 | de88c980-78f6-467c-a431-71bcc88e437f |

PR #60, opened 2018-11-19
config: unpanic when exoscale.toml is empty

greut:
:warning: master truncates the `~/.exoscale/exoscale.toml` file :warning:

**now**

```console
% truncate -s 0 ~/.exoscale/exoscale.toml
% exo config
default account not defined
```

**then**

```console
% truncate -s 0 ~/.exoscale/exoscale.toml
% exo config                 
We've found a "cloudstack.ini" configuration file with the following configurations:
...
```

PR #61, opened 2018-11-20
PR #57 broke credential reading from configuration

pyr:
This restores the functionality.
As an added bonus, this also restores `exo config` functionality, which also got broken in #57 (marshaling of configs failed both ways)

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -38,7 +38,7 @@ type account struct {
 	DefaultTemplate string
 }
 
-func (a account) GetSecret() string {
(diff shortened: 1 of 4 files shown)


greut:
cf #60 ?

pyr:
As far as I can tell, this is unrelated. #60 deals with config builds, here it's the unmarshal step of the configuration which breaks.

When unmarshaling TOML, private fields seem to be omitted, which made `secret` always be empty.

greut:
oh I see... snap. good catch

PR #62, opened 2018-11-20
config: use $XDG_CONFIG_HOME

pyr:
Configurations currently live in `~/.exoscale` this patchs switches the default location to `$XDG_CONFIG_HOME/exoscale` but keeps looking in the previous location to avoid breaking existing setups.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -170,7 +170,14 @@ func initConfig() {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pyr:
My use of `usr.HomeDir` is to mimic what is already being done (for instance, L189 and L190).
If you want to refactor this use fine, since it's orthogonal, it can be addressed later (especially since @pierre-emmanuelJ might have had a use case for `usr.HomeDir`?)

PR #69, opened 2018-11-29
Fix (#68) Api raw args not allowed

### LINKED ISSUE #68: api shouldn't allows raw args
Allowing any args in the API gives the false impression that they do something.

```console
% exo api vm list keyword=bar 
```

> 2018/11/28 09:17:18 GET /compute?apikey=EXO9d45a5d9bf68d5689a0eeb55&command=listVirtualMachines&expires=2018-11-28T09%3A27%3A18%2B0100&response=json&signature=...&signatureversion=3 HTTP/1.1

```console
% exo api vm list --keyword=bar
```

> 2018/11/28 09:18:14 GET /compute?apikey=EXO9d45a5d9bf68d5689a0eeb55&command=listVirtualMachines&expires=2018-11-28T09%3A28%3A14%2B0100&**keyword=bar**&response=json&signature=...signatureversion=3 HTTP/1.1

pierre-emmanuelJ:
go run main.go api zone list keyword=bar tags=tag
```
Raw arg(s) not allowed "keyword=bar, tags=tag"
Usage:
  exo api zone list [flags]

Aliases:
  list, ls, listZones

Flags:
      --available        true if you want to retrieve all available Zones. False if you only want to return the Zones from which you have at least one VM. Default is false. (default nil)
      --id UUID          the ID of the zone (default nil)
      --keyword string   List by keyword
      --name string      the name of the zone
      --page int
      --pagesize int
      --showcapacities   flag to display the capacity of the zones (default nil)
      --tags tag         List zones by resource tags (key/value pairs)
  -h, --help             help for list

Global Flags:
  -C, --config string        Specify an alternate config file [env EXOSCALE_CONFIG]
  -d, --debug                debug mode on
  -A, --use-account string   Account to use in config file [env EXOSCALE_ACCOUNT]
```
Fix #68

### DIFF
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -79,7 +79,7 @@ func buildCommands(methods []category) {
 			runCMD := func(cmd *cobra.Command, args []string) error {
 
 				if len(args) > 0 {
-					fmt.Fprintf(os.Stderr, "Raw arg(s) not allowed %q\n", strings.Join(args, ", ")) // nolint: errcheck
+					fmt.Fprintf(os.Stderr, "Raw arg(s) are not supported. Did you mean --%s ?\n", strings.Join(args, " --")) // nolint: errcheck
 					return cmd.Usage()
 				}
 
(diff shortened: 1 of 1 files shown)


greut:
my two cents.

```console
> exo api vm list foo=bar
Raw arguments are not supported. Did you mean?

exo api vm list --foo=bar
```

PR #78, opened 2018-12-19
Isos

greut:
Adding the isos command, but they won't show up in the exo api hierarchy.

```console
% exo api template
Template management

Usage:
  exo api template [command]

Aliases:
  template, temp

Available Commands:
  list        List all public, private, and privileged templates.

...

% exo api template listIsos | jq ".count"
6
```

### DIFF
--- a/cmd/api.go
+++ b/cmd/api.go
@@ -59,23 +59,6 @@ func buildCommands(methods []category) {
 				name = s.name
 			}
 
-			hiddenCMD := cobra.Command{
-				Use:    realName,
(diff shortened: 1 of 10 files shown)


pierre-emmanuelJ:
Added:

```
$
 exo  vm template list --iso
```

greut:
@pierre-emmanuelJ I've moved the `templates --isos` to #79 as @pyr don't want the isos side to be that visible. Hence the many changes in the layout file.

Issue #81, opened 2019-01-11
firewall add my-ip create two rules


greut:
Adding the SSH rule with my-ip create two rules when only the one containing my-ip should exist.

```console
% exo firewall show foo
(nothing)

% exo firewall add foo ssh --my-ip
INGRESS CIDR 0.0.0.0/0 tcp 22 (ssh)
        CIDR 80.224.112.21/32 tcp 22 (ssh)
```

PR #85, opened 2019-01-16
firewall: fix --my-ip and flagless add

falzm:
* if protocol is not set, rule will be created with
  * protocol set to all
  * port set to 0

```console
% exo firewall add egress --cidr 1.2.3.4/32 --port 45678
 SOURCE                   │ PROTOCOL │   PORT   │ 
 CIDR 1.2.3.4/32          │ all      │ 0        │             │ 
```

Maybe we should not accept --port without  --protocol 

* if protocol is set to all with --port

It is strange to have the possibility to set protocol to all with the cli but not through the portal

*exo 1.0.6 e886eca13deef7a2016a9c542fb3eca79deab4a4 (egoscale 0.13.2)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 -----
 
 - change: `firewall add` sets a CIDR by default
+- fix: `firewall add --my-ip` to not create the default CIDR
 
 1.0.4
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,26 @@
-1.0.5
+
+
+1.0.7
 -----
 
 - change: `firewall add` sets a CIDR by default
 - fix: `firewall add --my-ip` to not create the default CIDR
+
+1.0.6
+-----
+
+- feature: runstatus
+- feature: lab kube
+
+1.0.5
+-----
+
+- feature: sos recursive upload
+- feature: EXOSCALE_TRACE on the sos command
+- feature: allow secrets to come from an external source
+- feature: use XDG_CONFIG_HOME by default
+- feature: dns remove asks for confirmation
+- fix: `--my-ip` fix by @falzm
 
 1.0.4
 -----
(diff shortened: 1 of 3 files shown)


greut:
An empty call does something.

```console
% exo firewall add wide-open
```

|  TYPE   |     SOURCE     | PROTOCOL | PORT | DESCRIPTION |                  ID                  |
|---------|----------------|----------|------|-------------|--------------------------------------|
| INGRESS | CIDR 0.0.0.0/0 | all      | 0    |             | c5b01e37-6466-4cfd-8f2a-a4264da4df17 |

More doc has been added.

```console
% exo firewall add --help

firewall add --help

No arguments create a IPv4 wide open rule.

For ICMP rules, specify the icmp-code and icmp-type.

        firewall add <security group> --protocol icmp --icmp-type 8 --icmp-code 0

For the other protocols specify the port ranges.

        firewall add <security group> --protocol tcp --port 8000-8080

A set of predefined commands exists, such a ssh, ping or minecraft.

        firewall add <security group> ssh

Usage:
  exo firewall add <security group name | id>  [ssh | telnet | rdp | ...] (default preset rules) [flags]

```

What do you think?

pyr:
@greut creating a wide-open security group so easily seems a bit error-prone to me on first look

greut:
I happen to have one, but this is a different story.

|      NAME      |        DESCRIPTION        |                  ID                  |
|----------------|---------------------------|--------------------------------------|
| default        | Default Security Group    | 21a14a60-be48-4bde-8509-ae6ed0231ed0 |
| wide-open      | dummy (for test)          | 4bfe1073-a6d4-48bd-8f24-2ab586674092 |

falzm:
[ch355]

greut:
Closes #81 

@falzm right... sloppy changelog job :)

greut:
> As command is now explicit (wide-open) I see no concern in term of security/best practice

_wide-open_ is the name of the security group.

falzm:
Since the code change now has nothing to do with the original commit I suggest adapting the title and amending the commit message, or create a new PR.

PR #95, opened 2019-02-02
What do now?

jjasghar:
This to change it to a joke from It's Always Sunny in Philadelphia.

It would be a fun little joke for people to see.

Signed-off-by: JJ Asghar <jja@ibm.com>


greut:
@jjasghar cheers.

PR #112, opened 2019-02-28
lab: kube: add flag --version to create subcommand

### LINKED ISSUE #111: exo lab kube: add support for user-specified k8s version
Sometimes running `exo lab kube create` fails because the latest version of Kubernetes (retrieved at https://dl.k8s.io/release/stable.txt) is not yet packaged by Ubuntu). For example right now  the latest k8s version advertised is `1.13.4`, however creating exokube instances fails because the latest packaged version of the *kubelet/kubeadm/kubectl* utilities is `1.13.3`.

Adding an optional `--version` flag to the `exo lab kube create` command would allow users to specify a Kubernetes version (e.g. `1.13.3`) as a temporary workaround.

falzm:
Fixes #111.

### DIFF
--- a/cmd/kube_create.go
+++ b/cmd/kube_create.go
@@ -198,14 +198,13 @@ var kubeCreateCmd = &cobra.Command{
 			return err
 		}
 
-		kubernetesVersion, err := cmd.Flags().GetString("version")
+		requestedKubernetesVersion, err := cmd.Flags().GetString("version")
 		if err != nil {
 			return err
 		}
-		if kubernetesVersion == "" {
-			if kubernetesVersion, err = fetchKubernetesVersion(); err != nil {
-				return err
-			}
+		kubernetesVersion, err := fetchKubernetesVersion(requestedKubernetesVersion)
+		if err != nil {
+			return err
 		}
 
 		sizeOpt, err := cmd.Flags().GetString("size")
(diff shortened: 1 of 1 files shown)


cmd/kube_create.go
```diff
@@ -198,10 +198,15 @@ var kubeCreateCmd = &cobra.Command{
 			return err
 		}
 
-		kubernetesVersion, err := fetchKubernetesVersion()
+		kubernetesVersion, err := cmd.Flags().GetString("version")
```
pierre-emmanuelJ: I think can be cool to check if asking version exist, example:
https://github.com/kubernetes/kubernetes/releases/tag/v1.13.4
https://github.com/kubernetes/kubernetes/releases/tag/v{version}
https://github.com/kubernetes/kubernetes/releases/tag/{version}

greut: https://github.com/exoscale/cli/blob/742b3ae2000994b95557234591e038f9b0639b23/cmd/kube_create.go#L518-L544

falzm: Thanks @pierre-emmanuelJ & @greut, will do this.

falzm:
> LGTM if you have tested also it's good to me

Yep, tested OK 👍

dajac:
LGTM

PR #114, opened 2019-03-05
config: improve life of people without config

greut:
`exo config` should play nicer now.

**env**

```
% exo config     
remove ENV credentials variables to use config
% exo config show
                  environment variables

Account:          unknown
API Key:          EXO4d60e5e69f0225b35cb4fa05
API Secret:       ××××××××××

Default zone:     ch-dk-2
Default template: Linux Ubuntu 18.04 LTS 64-bit

Configuration:    environment variables
```

**no env**

```
% exo config
We've found a "cloudstack.ini" configuration file with the following configurations:
    ...

% exo config show
no accounts are defined
```

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -81,9 +81,9 @@ const (
 var configCmd = &cobra.Command{
 	Use:   "config",
 	Short: "Manage config for exo cli",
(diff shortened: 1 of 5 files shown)


pierre-emmanuelJ:
Zuper!

PR #129, opened 2019-04-11
Fix content-type sniffing on files < 512 bytes

### LINKED ISSUE #130: Content-Type sniffing broken on files < 512 bytes
Hello,

CLI doesn't sniff Content Type on files < 512bytes, leading to objects served with a "Content-Type: application/octet-stream".
Please find a PR that should fix the problem : #129 

Alex

adedommelin:
Fix #130


pierre-emmanuelJ:
Thanks a lot for your contribution, you catch a good point 😄
I'll try out this fix and come back to you ASAP

PR #135, opened 2019-04-24
startvirtualmachine: support the upcoming rescueprofile option

pyr:


cmd/vm_start.go
```diff
@@ -54,10 +63,16 @@ func startVirtualMachine(vmName string) error {
...
+	if (vmRescueProfile != nil) {
+		cmd.RescueProfile = *vmRescueProfile
+	}
+
+	_, err = asyncRequest(cmd, fmt.Sprintf("Starting %q ", vm.Name))
 	return err
 }
 
 func init() {
+	vmStartCmd.Flags().StringP("rescue-profile", "P", "", "option rescue profile when starting VM")
```
falzm: Captain Nitpick here: I'm not sure about the `-P` short flag. To me short flags should be reserved for global flags (e.g. help, config, verbose, version, account, debug...).

cmd/vm_start.go
```diff
@@ -18,9 +18,18 @@ var vmStartCmd = &cobra.Command{
 			return cmd.Usage()
 		}
 
+		var rescueProfileP *string
```
mcorbin: I think you don't have to define a pointer for this variable. You can pass `rescueProfile` directly to `startVirtualMachine` (like the `vmName` parameter). 
Because of the `omitempty` flag in the `StartVirtualMachine` call (here https://github.com/exoscale/egoscale/pull/401/files#diff-9d7e1c41fa657c6ea52f380e2d0c794dR318) if the string is empty the field will be discarded.

falzm:
Also, Travis is complaining about code formatting 🧐

cmd/vm_start.go
pyr: Thanks, I wasn't sure about the behavior

cmd/vm_start.go
pyr: can I prevent short flags by supplying an empty string here?

pyr: The doc says yes, I'll amend

falzm: Yes.

PR #160, opened 2019-06-24
sos: don't log upload activity by default

falzm:
This change disables SOS upload transfer logging activity, which now can
be enabled by passing the `--log|-l` flag.

### DIFF
--- a/cmd/sos_upload.go
+++ b/cmd/sos_upload.go
@@ -42,9 +42,21 @@ var sosUploadCmd = &cobra.Command{
 			return err
 		}
 
+		logfile, err := cmd.Flags().GetString("log")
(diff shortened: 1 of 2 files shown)


cmd/sos_upload.go
```diff
@@ -56,11 +68,6 @@ var sosUploadCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		lo, err := os.OpenFile("./log", os.O_RDWR|os.O_CREATE, os.ModePerm)
```
pyr: how about keeping it at `EXOSCALE_TRACE=1`?

pyr: sorry, misread

PR #163, opened 2019-07-12
firewall: check for CIDR or SG when adding a rule

falzm:
This change requires a CIDR or a Security Group to be specified when
adding a rule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ UNRELEASED
 ----------
 
 - Require a CIDR or a Security Group when adding a Security Group rule
+- Require protocol to be specified if a port is provided when adding a Security Group rule
 
 1.4.1
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 UNRELEASED
 ----------
 
-- Require a CIDR or a Security Group when adding a Security Group rule
 - Require protocol to be specified if a port is provided when adding a Security Group rule
 
 1.4.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+UNRELEASED
+----------
+
+- Require a CIDR or a Security Group when adding a Security Group rule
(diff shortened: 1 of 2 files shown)


jessicatoscani:
There is still a problem.
If protocol is not set, the new rule created is never what I was expecting

```
exo firewall add default --cidr 0.0.0.0/0 -P 8080 
┼──────────────────────────────────────┼─────────┼────────────────┼──────┼──────────┼─────────────┼
│                  ID                  │  TYPE   │     SOURCE     │ PORT │ PROTOCOL │ DESCRIPTION │
┼──────────────────────────────────────┼─────────┼────────────────┼──────┼──────────┼─────────────┼
│ 1aea8342-cd52-48fc-bf96-22d7c10661dd │ ingress │ CIDR 0.0.0.0/0 │ 0    │ all    
```

falzm:
@jessicatoscani I think a sensible fix would be to return an error if `--port` is specified without `--protocol`, are you OK with this?

jessicatoscani:
> @jessicatoscani I think a sensible fix would be to return an error if `--port` is specified without `--protocol`, are you OK with this?

Yes.

jessicatoscani:
Not sure about the mandatory --cidr

falzm:
> Not sure about the mandatory --cidr

It's not mandatory, but you either have to provide a CIDR or a Security Group.

falzm:
> > @jessicatoscani I think a sensible fix would be to return an error if `--port` is specified without `--protocol`, are you OK with this?
> 
> Yes.

Fixed.

jessicatoscani:
> > Not sure about the mandatory --cidr
> 
> It's not mandatory, but you either have to provide a CIDR or a Security Group.

Previous behaviour was fine:  add rule with  cidr = 0.0.0.0/0 when CIDR and SG were missing

falzm:
My bad, I've reverted my original fix and only kept the check on the port/protocol: @jessicatoscani can you confirm it's OK now?

PR #165, opened 2019-07-29
sos: fix `list` command panic

falzm:
This change prevents the `sos list` command from panicking if SOS
returns bogus entries.


falzm:
ch4852

falzm:
[ch4852]

PR #168, opened 2019-08-05
vm: create: check for maximum user-data length

falzm:
This change checks for maximum allowed user-data length before sending
the instance creation request:

```console
$ base64 -w0 userdata.txt | wc -c
32768

$ ./exo vm create test --cloud-init-file=userdata.txt
error: user-data maximum allowed length is 32768 bytes
```


greut:
I'd apply that check to the whole body, rather than the `user-data` part. Aka into egoscale.

PR #169, opened 2019-08-05
affinitygroup: fix delete confirmation prompt bug

falzm:
This change fixes a bug in the `affinitygroup delete` command, where user confirmation prompt reply would not have no effect and items would be deleted anyway.

### DIFF
--- a/cmd/affinitygroup_delete.go
+++ b/cmd/affinitygroup_delete.go
@@ -29,17 +29,16 @@ var affinitygroupDeleteCmd = &cobra.Command{
 				return err
 			}
 
-			tasks = append(tasks, task{
-				cmd,
-				fmt.Sprintf("delete %q affinity group", cmd.Name),
-			})
-
 			if !force {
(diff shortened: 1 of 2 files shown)


cmd/affinitygroup_delete.go
```diff
@@ -29,17 +29,16 @@ var affinitygroupDeleteCmd = &cobra.Command{
...
-
 			if !force {
 				if !askQuestion(fmt.Sprintf("sure you want to delete %q affinity group", arg)) {
 					continue
 				}
 			}
 
+			tasks = append(tasks, task{
+				cmd,
+				fmt.Sprintf("delete %q affinity group", cmd.Name),
```
chrisglass: Nit: "deleting" since the action is taking place.

Not blocking, it does fix the bug anyway, and we might as well be consistent (as you wish)

falzm: Good catch: fixed.

PR #174, opened 2019-08-08
Add support for building snap packages

chrisglass:
This branch makes a few changes to support running the exo binary from a
snap package (https://snapcraft.io).

The biggest difference is that snaps force the usage of $HOME/.exoscale/ as the source of configuration (a specific hole in the normal snap confinement mechanism was granted for this specific location).

Note: this is not meant to be "production ready" yet, a few rough
corners need to be polished first, but this at least allows manual
publication into the "edge channel".

Testing:

This is already published to the edge channel, so the following on any ubuntu system should just work:

```
sudo snap install --edge exoscale-cli
exo config
# Follow prompt
exo vm list
```

Note: a request for alias was made, so that in the future the installed binary can be "exo".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,6 @@ GO_FILES = $(shell find . -type f -name '*.go')
 .PHONY: all
 all: clean exo
 
-.PHONY: exo
 exo: $(GO_FILES)
 	go build ${GO_BUILDOPTS} -mod=vendor -o $@
 
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ GO_FILES = $(shell find . -type f -name '*.go')
 .PHONY: all
 all: clean exo
 
+.PHONY: exo
 exo: $(GO_FILES)
 	go build ${GO_BUILDOPTS} -mod=vendor -o $@
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ manpage
 contrib
 cli
 *.snap
+*.xdelta3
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -9,6 +9,7 @@ GO_FILES = $(shell find . -type f -name '*.go')
 .PHONY: all
 all: clean exo
 
+.PHONY: exo
```
pyr: It shouldn't be phony because it creates an executable called `exo`, should it?

chrisglass: Good point. Fixed.

greut:
Could the goreleaser config help? https://goreleaser.com/customization/#Snapcraft

falzm:
Good to know for the GoReleaser Snap support, thanks @greut.

chrisglass:
Yes, that is likely the best path forward since goreleaser can push to the snap store. Thanks!

In the mean time we need some paperwork with he snapstore to make sure the user experience is what we expect (read/write in $HOME/.exoscale, having an "exo" automatic alias). 

I guess this PR should have been a draft PR until then :)

falzm:
I'm a bit confused here: isn't one of the `.goreleaser.yml` or `snapcraft.yaml` files redundant, or are both needed? cc @pierre-emmanuelJ

chrisglass:
I guess the goreleaser is supposed to override the snapcraft file - but so far it's not been shown to actually work unfortunately.

Perhaps we could consider to keep the know-working snapcraft file to already publish snaps (by hand or otherwise, in a way that actually works), and consider migrating to goreleaser in another branch?

PR #184, opened 2019-09-15
Fix panic when EXOSCALE_ACCOUNT environment variable is set

7fELF:
There was a panic when the `EXOSCALE_ACCOUNT` environment variable was set. 

This was caused by a call to [FlagSet.Lookup](https://godoc.org/github.com/spf13/pflag#FlagSet.Lookup) which returned `nil` because the flag name was incorrect leading to a `nil pointer dereference`.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -119,13 +119,17 @@ var ignoreClientBuild = false
 func initConfig() {
 	envs := map[string]string{
 		"EXOSCALE_CONFIG":  "config",
-		"EXOSCALE_ACCOUNT": "account",
+		"EXOSCALE_ACCOUNT": "use-account",
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -119,13 +119,17 @@ var ignoreClientBuild = false
 func initConfig() {
 	envs := map[string]string{
 		"EXOSCALE_CONFIG":  "config",
-		"EXOSCALE_ACCOUNT": "account",
+		"EXOSCALE_ACCOUNT": "use-account",
```
pierre-emmanuelJ: 👍

falzm:
@7fELF don't forget to update the CHANGELOG file accordingly 🙏

PR #192, opened 2019-10-25
sos: fix TLS certificate issue on Windows

### LINKED ISSUE #189: Certificat error with exocli on Windows
Hi,

I get this error when I try to list my buckets from exocli.exe on Windows server.
```
C:\Users\Administrator\Downloads\exocli>exo.exe sos list
error: Get https://sos-ch-dk-2.exo.io/: x509: certificate signed by unknown authority
```
Regards,
Martial

falzm:
This change introduces a workaround to a bug in the Microsoft Windows
support in Go (https://github.com/golang/go/issues/16736). Some
heavily-chained certificates such as Exoscale SOS API's one are not
recognized as being signed by a trusted Certificate Authority.

Until the bug is fixed upstream, Windows users are required to use the
`--certs-file` flag with a path to a local file containing Exoscale SOS
API's certificate chain to the `exo sos` commands.

Fixes #189.

### DIFF
--- a/cmd/sos.go
+++ b/cmd/sos.go
@@ -1,40 +1,142 @@
 package cmd
 
 import (
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	minio "github.com/minio/minio-go/v6"
 	"github.com/spf13/cobra"
+	"golang.org/x/net/http2"
 )
 
 const (
-	sosZone = "ch-dk-2"
+	defaultSOSZone = "ch-dk-2"
 )
 
 // sosCmd represents the sos command
 var sosCmd = &cobra.Command{
 	Use:   "sos",
 	Short: "Simple Object Storage management",
+	Long: `Manage Exoscale Object Storage (SOS)
+
+IMPORTANT: Due to a bug in the Microsoft Windows support in the Go
+programming language (https://github.com/golang/go/issues/16736), some
+security certificates such as Exoscale SOS API's one are not recognized as
+being signed by a trusted Certificate Authority. In order to work around this
+issue until the bug is fixed upstream, Windows users are required to use the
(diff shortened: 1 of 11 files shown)


cmd/sos.go
```diff
@@ -1,40 +1,142 @@
...
+IMPORTANT: Due to a bug in the Microsoft Windows support in the Go
+programming language (https://github.com/golang/go/issues/16736), some
+security certificates such as Exoscale SOS API's one are not recognized as
+being signed by a trusted Certificate Authority. In order to work around this
+issue until the bug is fixed upstream, Windows users are required to use the
+"--certs-file" flag with a path to a local file containing Exoscale SOS API's
+certificate chain to the "exo sos" commands. This file can be downloaded from
+the following address:
+
+	https://www.exoscale.com/sos.crt
```
falzm: The actual address is up for debate, could be a link to the Community documentation. cc @llambiel @brutasse

falzm:
[ch6732]

cmd/sos_create.go
```diff
@@ -5,30 +5,27 @@ import (
...
 	"github.com/exoscale/egoscale"
-	minio "github.com/minio/minio-go/v6"
 	"github.com/spf13/cobra"
 )
 
 // createCmd represents the create command
 var sosCreateCmd = &cobra.Command{
 	Use:     "create <name>",
-	Short:   "create bucket",
+	Short:   "Create a bucket",
```
pierre-emmanuelJ: thx

falzm:
> LGTM, have you tested all refactored commands?

Yes, but feel free to do so on your end too :)

cmd/sos.go
exo-cedric: At some point (not so long in the future), we'll need some PKI-specific folder or hostname to distribute all PKI-related material (be it external, like the case at hand, or internal, if we choose to keep following that way).
(PS: this PKI material may be provided via an http**s**://... URI as long as there is no chicken-'n-egg problem)

falzm: IIRC the `exoscale.com` cert is universally trusted, only the `*.exo.io` causes issues.

PR #194, opened 2019-11-06
[sos acl add] Fix Full-Control on object when adding a manual Grant rule

pierre-emmanuelJ:
This PR fix the `exo sos acl add` command.
It's fix the `Full-Control` on object when adding a manual Grant rule, by taking the original object owner `ID` and not the current account `ID`

This PR is blocked until this PR: https://github.com/minio/minio-go/pull/1182 is Merged

Fix ch-6409


cmd/sos_acl.go
```diff
@@ -76,27 +78,27 @@ var sosAddACLCmd = &cobra.Command{
...
 
 		objInfo.Metadata.Add("content-type", objInfo.ContentType)
 
-		src := minio.NewSourceInfo(args[0], args[1], nil)
+		src := minio.NewSourceInfo(bucketName, objectName, nil)
 
-		_, okMeta := meta["X-Amz-Acl"]
-		_, okHeader := objInfo.Metadata["X-Amz-Acl"]
+		_, hasNewAmzACL := meta["X-Amz-Acl"]
+		_, hasAmzACL := objInfo.Metadata["X-Amz-Acl"]
```
falzm: Please comment what those variables are needed for: 2 variables with one named "new" is not enough to understand what is going on here.

cmd/sos_acl.go
```diff
@@ -31,7 +31,7 @@ const (
 
 // aclCmd represents the acl command
 var sosACLCmd = &cobra.Command{
-	Use:   "acl <bucket name> <object name> [object name] ...",
```
falzm: 👍

cmd/sos_acl.go
pierre-emmanuelJ: 👍

PR #195, opened 2019-11-09
bug: adding a restricted key to config

illi-j:
This PR removes the function of being able to add an API key in the config file after its creation & adds the function of adding a restricted API key in the config file with the command `exo config`

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -199,8 +199,14 @@ func getAccount() (*account, error) {
 		Secret:   "",
 	}
 
+	isRestricted := false
+
+	if askQuestion("is it a restricted account?") {
+		isRestricted = true
+	}
+
 	for i := 0; ; i++ {
-		if i > 0 {
+		if i > 0 || isRestricted {
 			endpoint, err := readInput(reader, "API Endpoint", account.Endpoint)
 			if err != nil {
 				return nil, err
@@ -236,6 +242,19 @@ func getAccount() (*account, error) {
 		fmt.Printf("Checking the credentials of %q...", account.Key)
 		resp, err := client.GetWithContext(gContext, egoscale.Account{})
 		if err != nil {
+			if isRestricted {
+				fmt.Print(" can't get account!\n\n")
+				acc, err := readInput(reader, "Account", account.Account)
+				if err != nil {
+					return nil, err
+				}
+				if acc != "" {
+					account.Account = acc
+				}
+
+				break
+			}
+
 			fmt.Print(` failure.
 
 Let's start over.
@@ -272,12 +291,21 @@ Let's start over.
 		account.Name = name
 	}
 
-	defaultZone, err := chooseZone(account.Name, client)
+	account.DefaultZone, err = chooseZone(account.Name, client)
 	if err != nil {
-		return nil, err
+		if isRestricted {
+			defaultZone, err := readInput(reader, "Zone", account.DefaultZone)
+			if err != nil {
+				return nil, err
+			}
+			if defaultZone != "" {
+				account.DefaultZone = defaultZone
+			}
+		} else {
+			return nil, err
+		}
 	}
 
-	account.DefaultZone = defaultZone
 	account.DNSEndpoint = strings.Replace(account.Endpoint, "/compute", "/dns", 1)
 
 	return account, nil
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -199,14 +199,8 @@ func getAccount() (*account, error) {
 		Secret:   "",
 	}
 
-	isRestricted := false
-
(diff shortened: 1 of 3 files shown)


illi-j:
We have another problem, this pr fixes the problem when we add the key in the config file when we create a new key but if the person uses the exo config command to add this key he will have the same problem.
And I don't think removing the check credentials from this order is the right solution.

falzm:
@illi-j sorry I haven't been clearer in my previous comment: the CLI shouldn't even ask the user if he/she wants to add an API key created using `exo iam apikey create`:

```
$ exo lab iam apikey create list-zones -o compute/listZones
┼────────────┼─────────────────────────────────────────────┼
│ Name       │ list-zones                                  │
│ Key        │ EXOd2227f10c642539f965f552a                 │
│ Secret     │ vz6OajkNpwRVgCfpIU6M8D1dazUuo1bzOyC3AHJAlK0 │
│ Operations │ compute/listZones                           │
│ Type       │ restricted                                  │
┼────────────┼─────────────────────────────────────────────┼
[+] do you wish to add this account in your config file? [yN]:
```

cmd/config.go
```diff
@@ -236,6 +242,19 @@ func getAccount() (*account, error) {
 		fmt.Printf("Checking the credentials of %q...", account.Key)
 		resp, err := client.GetWithContext(gContext, egoscale.Account{})
 		if err != nil {
+			if isRestricted {
+				fmt.Print(" can't get account!\n\n")
+				acc, err := readInput(reader, "Account", account.Account)
```
7fELF: Can this input read be more explicit? What do you expect from the user here?

cmd/config.go
```diff
@@ -272,12 +291,21 @@ Let's start over.
...
-	defaultZone, err := chooseZone(account.Name, client)
+	account.DefaultZone, err = chooseZone(account.Name, client)
 	if err != nil {
-		return nil, err
+		if isRestricted {
+			defaultZone, err := readInput(reader, "Zone", account.DefaultZone)
+			if err != nil {
+				return nil, err
+			}
+			if defaultZone != "" {
```
7fELF: What happens if it is empty?

cmd/config.go
```diff
@@ -236,6 +242,26 @@ func getAccount() (*account, error) {
 		fmt.Printf("Checking the credentials of %q...", account.Key)
 		resp, err := client.GetWithContext(gContext, egoscale.Account{})
 		if err != nil {
+			if isRestricted {
+				fmt.Print(` can't get account!
+
+Please enter your account name.
+
+				`)
```
pierre-emmanuelJ: I think you can remove white space from this line

cmd/config.go
```diff
@@ -199,8 +199,14 @@ func getAccount() (*account, error) {
 		Secret:   "",
 	}
 
+	isRestricted := false
+
+	if askQuestion("is it a restricted account?") {
```
falzm: What's a "restricted account"?

illi-j: It is an account whose API key has been created with IAM and which is of the "restricted" type

7fELF: We can avoid the question with something like this https://github.com/exoscale/cli/compare/iam/ch7024...7fELF/iam/ch7024

falzm: Thank you @7fELF, I was about give a similar answer: ideally, we'd even identify the error at egoscale level and return an `egoscale.ErrorPermissionDenied` error that can be matched at client level and return an appropriate error message without asking questions...

illi-j: The fact of asking the question is not the problem, the problem was to find the best way to solve the bug without removing, as it was your idea, to remove code that for me seems useful.
And yes, thank you @7fELF

cmd/config.go
```diff
@@ -236,6 +236,22 @@ func getAccount() (*account, error) {
 		fmt.Printf("Checking the credentials of %q...", account.Key)
 		resp, err := client.GetWithContext(gContext, egoscale.Account{})
 		if err != nil {
+			if egoerr, ok := err.(*egoscale.ErrorResponse); ok && egoerr.ErrorCode == egoscale.ErrorCode(403) {
+				fmt.Println("This key is restricted, Please enter your account name.")
```
falzm: This error message is confusing: the user doesn't necessary know what is the meaning of *restricted* and what is has to do with the fact that we're asking him/her to provide account information.

You should replace `Checking the credentials of %q...` (because we're actually doing more than just checking the credentials) with `Retrieving account information...`, then if the specified API key is indeed restricted and doesn't allow the CLI to do it then return a message ` failed. Please enter your account information:`, or simply ` done` if the account information has been successfully retrieved.

illi-j: I agree with you 👍

PR #196, opened 2019-11-13
Show Private Networks in "exo vm show" (ch-6418)

pierre-emmanuelJ:
Currently the `exo vm show` command doesn't show the Private Networks the instance is attached to, although this information is show in the portal equivalent view.

### DIFF
--- a/cmd/vm_show.go
+++ b/cmd/vm_show.go
@@ -90,7 +90,7 @@ func showVM(name string) (outputter, error) {
 		SSHKey:             vm.KeyPair,
 		SecurityGroups:     make([]string, len(vm.SecurityGroup)),
 		AntiAffinityGroups: make([]string, len(vm.AffinityGroup)),
-		PrivateNetworks:    make([]string, len(vm.Nic)),
+		PrivateNetworks:    make([]string, 0, len(vm.Nic)-1),
 	}
 
 	for i, sg := range vm.SecurityGroup {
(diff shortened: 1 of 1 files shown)


cmd/vm_show.go
```diff
@@ -99,6 +101,14 @@ func showVM(name string) (outputter, error) {
 		out.AntiAffinityGroups[i] = aag.Name
 	}
 
+	for _, nic := range vm.Nic {
+		if nic.ID == vm.DefaultNic().ID {
+			continue
+		}
```
greut: It seems to be a convoluted way of doing.
```golang
if nic.IsDefault {
    continue
}
```

https://github.com/exoscale/egoscale/blob/master/virtual_machines.go#L146

pierre-emmanuelJ: Thanks good catch! I did not see that 😉

cmd/vm_show.go
```diff
@@ -89,6 +90,7 @@ func showVM(name string) (outputter, error) {
 		SSHKey:             vm.KeyPair,
 		SecurityGroups:     make([]string, len(vm.SecurityGroup)),
 		AntiAffinityGroups: make([]string, len(vm.AffinityGroup)),
+		PrivateNetworks:    make([]string, 0, len(vm.Nic)-1),
```
falzm: No need to mess with slice capacity here, either create an empty slice with a 0 length and use `append()` to add elements to it, or create it with final length (`make([]string, len(vm.Nic)-1))` and assign elements using the `for i, nic := range vm.Nic { ... }` position index (`out.PrivateNetworks[i] = vm.Nic[i]`).

pierre-emmanuelJ: Why not

PR #202, opened 2019-11-22
config: add a dedicated account creation command

falzm:
This change introduces an explicit `exo config add` command allowing
users to create a CLI configuration accounts. Includes some light
internal refactoring to make code more readable.

```console
$ ./exo config add blah
[+] API Key [none]: EXOxxxxxxxxxxxxxxxxxxxxxxx
[+] Secret Key [none]: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Retrieving account information... done!

[+] Name [xxxxxx@exoscale.com]: blah
✔ ch-gva-2
You chose "ch-gva-2"
[+] Set [blah] as default account? [yN]: y

$ ./exo config
To configure a new account, run `exo config add`
Use the arrow keys to navigate: ↓ ↑ → ←
? Configured accounts (* = default account):
  ▸ alice
    bob
    carl
    blah*
```

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -87,39 +87,48 @@ const (
 // configCmd represents the config command
 var configCmd = &cobra.Command{
 	Use:   "config",
-	Short: "Manage config for exo CLI",
+	Short: "Exoscale CLI configuration management",
 	RunE:  configCmdRun,
 }
 
 func configCmdRun(cmd *cobra.Command, args []string) error {
+	var defaultAccountMark = "*"
+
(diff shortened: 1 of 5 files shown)


falzm:
> Is the `exo config set` command really useful?
> because we can already change the default account with the command `exo config`

@illi-j one usage I can think of is in non-interactive situations (e.g. scripts), but I agree it's not extremely useful and could be removed in an upcoming major version of the CLI.

PR #204, opened 2019-12-02
add sos show command

illi-j:
This PR add a new SOS command.

- `sos show`

```
%exo sos show
Usage:
  exo sos show <bucket name> <oject name> [flags]

Aliases:
  show, get

Flags:
  -h, --help   help for show

Global Flags:
      --certs-file string        Path to file containing additional SOS API X.509 certificates
  -O, --output-format string     Output format (table|json|text)
  -Q, --quiet                    Quiet mode (disable non-essential command output
```

```
%exo sos show test T-DOP-600_ansible.pdf
┼───────────────────────┼─────────────────────────────────────────────────────────────┼
│ T-DOP-600 ANSIBLE PDF │                                                             │
┼───────────────────────┼─────────────────────────────────────────────────────────────┼
│ ACL                   │   ACCESS    VALUE                                           │
│                       │   Canned   private                                          │
│                       │                                                             │
│ Metadata              │   KEY   VALUE                                               │
│                       │   dc    dcv                                                 │
│                       │   dv    dv                                                  │
│                       │                                                             │
│ Headers               │       KEY             VALUE                                 │
│                       │   content-type   application/pdf                            │
│                       │                                                             │
│ URL                   │ https://sos-ch-gva-2.exo.io/test/T-DOP-600_ansible.pdf      │
┼───────────────────────┼─────────────────────────────────────────────────────────────┼
```

### DIFF
--- a/cmd/sos_show.go
+++ b/cmd/sos_show.go
@@ -45,7 +45,6 @@ func (o *sosShowOutput) toTable() {
 
 		switch {
 		case o.ACL[0].Access == "Canned":
-			at.SetHeader([]string{"canned"})
 			at.Append([]string{o.ACL[0].Value})
 		default:
 			at.SetHeader([]string{"access", "value"})
--- a/cmd/sos_show.go
+++ b/cmd/sos_show.go
@@ -26,7 +26,6 @@ type sosHeadersShowOutput struct {
 }
 
 type sosShowOutput struct {
-	Object   string                  `json:"object"`
 	URL      string                  `json:"url"`
 	ACL      []sosACLShowOutput      `json:"acl"`
 	Metadata []sosMetadataShowOutput `json:"metadata"`
@@ -39,16 +38,25 @@ func (o *sosShowOutput) toText() { outputText(o) }
 
 func (o *sosShowOutput) toTable() {
 	t := table.NewTable(os.Stdout)
-	t.SetHeader([]string{o.Object})
 
 	if o.ACL != nil {
 		buf := bytes.NewBuffer(nil)
 		at := table.NewEmbeddedTable(buf)
-		at.SetHeader([]string{"access", "value"})
-		for _, a := range o.ACL {
-			at.Append([]string{a.Access, a.Value})
+
+		switch {
+		case o.ACL[0].Access == "Canned":
+			at.SetHeader([]string{"canned"})
+			at.Append([]string{o.ACL[0].Value})
+		default:
+			at.SetHeader([]string{"access", "value"})
+
+			for _, a := range o.ACL {
+				at.Append([]string{a.Access, a.Value})
+			}
 		}
+
 		at.Render()
+
 		t.Append([]string{"ACL", buf.String()})
 	}
 
@@ -56,10 +64,13 @@ func (o *sosShowOutput) toTable() {
 		buf := bytes.NewBuffer(nil)
 		mt := table.NewEmbeddedTable(buf)
 		mt.SetHeader([]string{"key", "value"})
+
 		for _, m := range o.Metadata {
 			mt.Append([]string{m.Key, m.Value})
 		}
+
 		mt.Render()
+
 		t.Append([]string{"Metadata", buf.String()})
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/sos_show.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	t.Append([]string{"URL", fmt.Sprint(o.URL)})
+
+	t.Render()
+}
+
+// sosShowCmd represents the show command
+var sosShowCmd = &cobra.Command{
+	Use:     "show <bucket name> <oject name>",
```
falzm:
```suggestion
	Use:     "show <bucket name> <object name>",
```

cmd/sos_show.go
```diff
@@ -0,0 +1,187 @@
...
+	Headers  []sosHeadersShowOutput  `json:"headers"`
+}
+
+func (o *sosShowOutput) toJSON() { outputJSON(o) }
+
+func (o *sosShowOutput) toText() { outputText(o) }
+
+func (o *sosShowOutput) toTable() {
+	t := table.NewTable(os.Stdout)
+	t.SetHeader([]string{o.Object})
```
falzm: I'm not sure we should add a header with the object name:

* it can result in a veeeeeeeeeeeeeeeeeeeeeery wide column if the object name is veeeeeeeeeeeeeeeeeeeeeery long
* according to your example, the displayed object is not exactly the actual object name (underscores are replaced with whitespaces, and so will be other characters I guess...)

Thus, I suggest not putting a table header and instead add an `Object | <the object name>` row in the table.

cmd/sos_show.go
```diff
@@ -0,0 +1,187 @@
...
+					Key:   k,
+					Value: v[0],
+				})
+			}
+		}
+	}
+
+	out := sosShowOutput{
+		Object:   object,
+		URL:      fmt.Sprintf("https://sos-%v.exo.io/%v/%v", location, bucket, object),
```
pierre-emmanuelJ: Can also be `%s` ;)

cmd/sos_show.go
```diff
@@ -0,0 +1,187 @@
...
+func (o *sosShowOutput) toText() { outputText(o) }
+
+func (o *sosShowOutput) toTable() {
+	t := table.NewTable(os.Stdout)
+	t.SetHeader([]string{o.Object})
+
+	if o.ACL != nil {
+		buf := bytes.NewBuffer(nil)
+		at := table.NewEmbeddedTable(buf)
+		at.SetHeader([]string{"access", "value"})
```
falzm: I'm not sure this format is suitable for canned ACLs: when an object is applied a canned ACL, I suggest simplifying the output to the canned ACL name only, and switch to "detailed" table-based output only if the object has complex manually set ACLs.

cmd/sos_show.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	cannedACL, okHeader := objInfo.Metadata["X-Amz-Acl"]
+
+	t := table.NewTable(os.Stdout)
+	t.SetHeader([]string{"ACL", "METADATA", "HEADER"})
+
+	var (
+		acls      []sosACLShowOutput
+		metadatas []sosMetadataShowOutput
```
falzm: *Metadata* is singular.

```suggestion
		metadata []sosMetadataShowOutput
```

cmd/sos_show.go
```diff
@@ -0,0 +1,187 @@
...
+
+	objInfo, err := sosClient.GetObjectACL(bucket, object)
+	if err != nil {
+		return nil, err
+	}
+
+	cannedACL, okHeader := objInfo.Metadata["X-Amz-Acl"]
+
+	t := table.NewTable(os.Stdout)
+	t.SetHeader([]string{"ACL", "METADATA", "HEADER"})
```
falzm: Dead code?

cmd/sos_show.go
```diff
@@ -0,0 +1,187 @@
...
+		if len(v) > 0 {
+			if isGrantACL(k) {
+				s := getGrantValue(v)
+				acls = append(acls, sosACLShowOutput{
+					Access: formatGrantKey(k),
+					Value:  s,
+				})
+			}
+
+			k = strings.ToLower(k)
```
pierre-emmanuelJ: I think the `ToLower` here  will erase the key if the user want to have an upper case Key to the header.
Same problem here: https://github.com/exoscale/cli/blob/master/cmd/sos_metadata.go#L186

cmd/sos_show.go
illi-j: I think we don't even need to display it

falzm: Yes, that was my point 😉

cmd/sos_show.go
```diff
@@ -0,0 +1,197 @@
...
+
+func (o *sosShowOutput) toTable() {
+	t := table.NewTable(os.Stdout)
+
+	if o.ACL != nil {
+		buf := bytes.NewBuffer(nil)
+		at := table.NewEmbeddedTable(buf)
+
+		switch {
+		case o.ACL[0].Access == "Canned":
```
falzm: I was thinking of an even simpler display for canned ACLs, something like:

```
./exo sos show xxxx-xxxx IMG_1184
┼─────────┼────────────────────────────────────────────────────┼
│ ACL     │ public-read                                        │
│         │                                                    │
│ Headers │       KEY          VALUE                           │
│         │   content-type   image/jpeg                        │
│         │                                                    │
│ URL     │ https://sos-de-fra-1.exo.io/xxxx-xxxx/IMG_1184.jpg │
┼─────────┼────────────────────────────────────────────────────┼
```

WDYT?

illi-j: I think it's good to tell him what it is, but, it's true that on your example it's more readable

PR #206, opened 2019-12-02
Remove NetworkOffering to create a Private Network

pierre-emmanuelJ:
Exoscale API has changed and now no need to use NetworkOffering to create a Private-Network


vendor/modules.txt
```diff
@@ -54,7 +54,7 @@ github.com/dlclark/regexp2
 github.com/dlclark/regexp2/syntax
 # github.com/dustin/go-humanize v1.0.0
 github.com/dustin/go-humanize
-# github.com/exoscale/egoscale v0.20.1
+# github.com/exoscale/egoscale v0.22.0
```
chrisglass: :+1:

PR #217, opened 2019-12-17
[sos] remove the listZones call

illi-j:
This PR removed the listZones call in `exo sos create` & used the account defaultZone (if it possible) in newSOSClient


pierre-emmanuelJ:
👍

Issue #197, opened 2019-11-13
exoscale.toml created in the wrong directory on Windows


janoszen:
When running `exo config` on Windows, the file in `C:\Users\USERNAME\.config\exoscale\exoscale.toml` is created with 0 bytes, and the actual `exoscale.toml` is created with the correct content in the current directory.

pierre-emmanuelJ:
Thanks to post this issue 👍
We will try to reproduce the bug to solve it

janoszen:
This issue seems to be fixed in git master on Windows 10 Pro.

pierre-emmanuelJ:
Thank you!
Solved by: https://github.com/exoscale/cli/pull/193

pierre-emmanuelJ:
On my side tested on Windows server 2019

horakg:
looks like that bug still exists, just extracted exo cli 1.10.0 on my windows notebook to C:\users\mk6s56\mydrive\exo and did an "exo config" afterwards

C:\Users\mk6s56\myDrive\exo>exo config
No Exoscale CLI configuration found

Hi happy Exoscalian, some configuration is required to use exo.

We now need some very important information, find them there.
        <https://portal.exoscale.com/account/profile/api>

[+] API Key [none]: myapikey
[+] Secret Key [none]: mysecretkey
Retrieving account information... done!

[+] Name [a1-digital]:
v at-vie-1

exo vm list - is not possible as it complains about a missing configuration

C:\Users\mk6s56\myDrive\exo>exo vm list
no accounts were found into "C:\\Users\\mk6s56\\.config\\exoscale\\exoscale.toml"

If I manual select the correct exoscale.toml config it works
C:\Users\mk6s56\myDrive\exo>exo -C exoscale.toml vm list
┼──────────────────────────────────────┼─────────────────────────────────────────┼─────────────┼──────────┼─────────┼─────────────────┼
│                  ID                  │                  NAME                   │    SIZE     │   ZONE   │  STATE  │   IP ADDRESS    │
┼──────────────────────────────────────┼─────────────────────────────────────────┼─────────────┼──────────┼─────────┼─────────────────┼
│ a63888b9-2bec-4c1f-b9bf-f90857a8d744 │ westallianz-pentest                     │ Extra-large │ at-vie-1 │ Running │ 194.182.174.211 │
[...]

PR #220, opened 2020-02-03
use go 1.13 and rely on os.UserConfigDir to find config location

pyr:
This should fix failures to find the config file's location on
Windows. As per a recommendation from @greut, thanks!

Somewhat related to #197 though it might not fix the issue.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -200,18 +200,11 @@ func initConfig() {
 		}
 	}
 
-	xdgHome, found := os.LookupEnv("XDG_CONFIG_HOME")
-	if found {
-		gConfigFolder = path.Join(xdgHome, "exoscale")
-	} else {
-		home, found := os.LookupEnv("HOME")
-		if found {
-			gConfigFolder = path.Join(home, ".config", "exoscale")
-		} else {
-			// The XDG spec specifies a default XDG_CONFIG_HOME in $HOME/.config
-			gConfigFolder = path.Join(usr.HomeDir, ".config", "exoscale")
-		}
+	cfgdir, err := os.UserConfigDir()
+	if err != nil {
+		log.Fatalf("could not find configuration directory: %s", err)
 	}
+	gConfigFolder = path.Join(cfgdir, "exoscale")
 
 	// Snap packages use $HOME/.exoscale (as negotiated with the snap store)
 	if _, snap := os.LookupEnv("SNAP_USER_COMMON"); snap {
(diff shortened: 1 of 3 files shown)


pyr:
Naive fix attempt, builds locally. I don't know the implications of requiring go 1.13 though.

pyr:
@horakg can you try a build off of this branch to see if the behavior changes?

cmd/root.go
```diff
@@ -200,18 +200,12 @@ func initConfig() {
...
-		if found {
-			gConfigFolder = path.Join(home, ".config", "exoscale")
-		} else {
-			// The XDG spec specifies a default XDG_CONFIG_HOME in $HOME/.config
-			gConfigFolder = path.Join(usr.HomeDir, ".config", "exoscale")
-		}
+	cfgdir, err := os.UserConfigDir()
+	if err != nil {
+		log.Fatal(fmt.Errorf("could not find configuration directory: %s", err))
+		return
```
7fELF:
```suggestion
```
[`log.Fatal`](https://godoc.org/log#Logger.Fatal) calls calls `os.Exit(1)`, no need to return

cmd/root.go
```diff
@@ -200,18 +200,12 @@ func initConfig() {
...
-		home, found := os.LookupEnv("HOME")
-		if found {
-			gConfigFolder = path.Join(home, ".config", "exoscale")
-		} else {
-			// The XDG spec specifies a default XDG_CONFIG_HOME in $HOME/.config
-			gConfigFolder = path.Join(usr.HomeDir, ".config", "exoscale")
-		}
+	cfgdir, err := os.UserConfigDir()
+	if err != nil {
+		log.Fatal(fmt.Errorf("could not find configuration directory: %s", err))
```
7fELF:
```suggestion
		log.Fatalf("could not find configuration directory: %s", err)
```

horakg: @pyr how can I build it for windows?

falzm:
@horakg here is a Windows build of this branch: 
[exoscale-cli_v1.10.0-SNAPSHOT-7b2f70e_windows_amd64.zip](https://github.com/exoscale/cli/files/4151927/exoscale-cli_v1.10.0-SNAPSHOT-7b2f70e_windows_amd64.zip)

horakg:
@falzm @pyr this version works on my office pc
Shall I also try it at home?

falzm:
@horakg for belt and suspenders' sake I'd say yes: if confirmed, we can definitively put this problem behind :)

horakg:
works also at home

falzm:
Great, thanks for confirming.

PR #221, opened 2020-02-07
config: improve account listing UX

falzm:
This change improves the UX of listing configured accounts and adds a
new shortcut to create a new one directly from the list.

Also fixes a bug where a non-existent account name would not be detected
properly.

```console
$ ./exo config
Use the arrow keys to navigate: ↓ ↑ → ←
? Configured accounts (* = default account):
  ▸ alice*
    bob
    <Configure a new account>
```

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -95,7 +95,7 @@ var configCmd = &cobra.Command{
 
 func configCmdRun(cmd *cobra.Command, args []string) error {
 	var (
-		defaultAccountMark = promptui.Styler(promptui.FGBold, promptui.FGYellow)("*")
+		defaultAccountMark = promptui.Styler(promptui.FGYellow)("*")
 		newAccountLabel    = "<Configure a new account>"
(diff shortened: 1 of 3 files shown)


cmd/config.go
```diff
@@ -94,18 +94,22 @@ var configCmd = &cobra.Command{
 }
 
 func configCmdRun(cmd *cobra.Command, args []string) error {
-	var defaultAccountMark = "*"
+	var (
+		defaultAccountMark = promptui.Styler(promptui.FGBold, promptui.FGYellow)("*")
```
illi-j:
```suggestion
		defaultAccountMark = promptui.Styler(promptui.FGYellow)("*")
```
When you put it in bold, it doesn't show the color.
I suggest you put either bold or yellow.

Screenshot 2020-02-10 at 11 59 15

falzm: Good catch, thanks :)

PR #226, opened 2020-03-03
Use non EOL Ubuntu as based distro

mcanevet:
As Ubuntu 18.10 is EOL, we can't install packages using exoscale/cli as base image because Ubuntu mirrors are not available anymore.


mcanevet:
Thanks @falzm, do you plan to release a new image anytime soon (even with latest tag)? or should I wait for v1.11.0 to be released?

falzm:
@mcanevet `exoscale/cli:latest` has just been pushed :)

mcanevet:
Great! thanks

PR #227, opened 2020-03-12
Use vendorized source to generate documentation

zyegfryed:
This way the documentation stays in sync with the code.

### DIFF
--- a/website/prepare.sh
+++ b/website/prepare.sh
@@ -5,7 +5,7 @@ cp ../README.md content/_index.md
 mkdir -p static
 
 cd ..
-go run doc/main.go
+go run -mod vendor doc/main.go
 
 set +xe
 echo "we are now ready to: hugo serve"


falzm:
Just out of curiosity, what are the effects of this change? It's not possible to see the difference with git.

zyegfryed:
> Just out of curiosity, what are the effects of this change? It's not possible to see the difference with git.

First, a bit of context. I was trying to generate the `exo cli` documentation and noticed that a modification in the vendorized source code was **not** reflected in the generated documentation.
I also noticed that most (if not all) the commands in the `Makefile` were using the `-mod=vendor` flag, i.e. referring to the vendorized source code. I added 1+1 and end-up to that PR.

Now, to answer your question, the effect of this change is to use the vendorized source code instead of the latest version of a dependency as the input source for the generated documentation. I.e. the vendorized source code becomes the source of truth for either the generated executable and the documentation it relies on.

falzm:
> Now, to answer your question, the effect of this change is to use the vendorized source code instead of the latest version of a dependency as the input source for the generated documentation. I.e. the vendorized source code becomes the source of truth for either the generated executable and the documentation it relies on.

Yes I could see that 😄 But I've tried generating the website files with and without `-mod vendor` and could not see any difference in the output...

zyegfryed:
> Yes I could see that 😄 But I've tried generating the website files with and without `-mod vendor` and could not see any difference in the output...

Try cleaning the `website/content` folder first with `rm website/content/*.md`, then edit a file. 

Let's say we modified `vendor/github.com/exoscale/egoscale/virtual_machines.go` like the following:

```diff

-       Tags              []ResourceTag `json:"tags,omitempty" doc:"List resources by tags (key/value pairs)"`
+       Tags              []ResourceTag `json:"tags,omitempty" doc:"List resources by tags (key/value pairs). BOOOOOOOOOM!"
```

Generate the website again with `./prepare.sh` and run `hugo serve`. Browse to http://localhost:1313/cli/exo_api_vm_list/ and you should see "BOOOOOOOOOM!" in the `--tags` content.

pierre-emmanuelJ:
Thanks @zyegfryed 

for example `exo api`command (usage, description...etc) is based on egoscale package.
true @zyegfryed,  so if egoscale in vendor is not the same version as mod version so the doc generation can be different.

zyegfryed:
Cheers 🍰

PR #228, opened 2020-03-16
template register: support output customization

falzm:
This change adds support for `vm template register` command output
customization via the global `--output-format` flag. Also, async
commands output support messages (e.g. command waiter message &
termination status) are now written to stderr instead of stdout so
that they can be discarded independently from standard commands
output.

### DIFF
--- a/cmd/template_register.go
+++ b/cmd/template_register.go
@@ -2,17 +2,20 @@ package cmd
 
 import (
 	"fmt"
-	"os"
+	"strings"
 
-	"github.com/exoscale/cli/table"
 	"github.com/exoscale/egoscale"
 	"github.com/spf13/cobra"
 )
 
 // templateRegisterCmd registers a template
 var templateRegisterCmd = &cobra.Command{
-	Use:     "register",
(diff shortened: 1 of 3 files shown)


cmd/template_register.go
```diff
@@ -80,34 +83,33 @@ var templateRegisterCmd = &cobra.Command{
...
-	resp, err := asyncRequest(registerTemplate, fmt.Sprintf("Registering the template"))
+	resp, err := asyncRequest(registerTemplate, "Registering the template")
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	templates := resp.(*[]egoscale.Template)
+	if len(*templates) != 1 {
+		return nil, nil
```
pierre-emmanuelJ: Is it normal?

falzm: My guess is that the `registerTemplate` call should not return a response containing either 0 template or more than one, so if it happens we're in a unknown situation anyhow... Best not do anything ¯\\\_(ツ)_/¯ (any known error has been returned before this point)

PR #231, opened 2020-04-01
iam: improve apikey commands output UX

falzm:
This change improves the UX of the `iam apikey` commands:

- Sort operations list in alphabetical order
- Don't display the operations in the `apikey list` view, as a long list
  of operations produces ugly results

### DIFF
--- a/cmd/iam_apikey.go
+++ b/cmd/iam_apikey.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/exoscale/egoscale"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cmd/iam_apikey.go
```diff
@@ -50,6 +51,8 @@ func getAPIKeyByName(name string) (*egoscale.APIKey, error) {
 		return nil, fmt.Errorf(`more than one element found: %d`, count)
 	}
 
+	sort.Strings(apiKeys[0].Operations)
```
jeromer, line 54: What if `apiKeys` is empty ?

falzm: At this point it cannot be, this has been checked by the `switch` statement a few lines above.

greut: it fails at line 49.

PR #232, opened 2020-04-09
Add: show account limits

illi-j:
This command list  the safety limits currently enforced on your account.


illi-j:
``` shell
exo limits -h
This command lists the safety limits currently enforced on your account.

Supported output template annotations: .Resource, .Used, .Available, .Max

Usage:
  exo limits [flags]

Flags:
  -h, --help   help for limits

Global Flags:
  -Q, --quiet                    Quiet mode (disable non-essential command output
```

illi-j:
``` shell
%> exo limits
┼──────────────────┼──────┼───────────┼─────┼
│     RESOURCE     │ USED │ AVAILABLE │ MAX │
┼──────────────────┼──────┼───────────┼─────┼
│ Instances        │ 3    │ 17        │ 20  │
│ Snapshots        │ 0    │ 30        │ 30  │
│ Templates        │ 0    │ 10        │ 10  │
│ Private Networks │ 0    │ 32        │ 32  │
│ IP Addresses     │ 0    │ 20        │ 20  │
┼──────────────────┼──────┼───────────┼─────┼
```

cmd/limits.go
```diff
@@ -0,0 +1,119 @@
...
+	"strings"
+
+	"github.com/exoscale/egoscale"
+	"github.com/spf13/cobra"
+)
+
+type LimitsItemOutput struct {
+	Resource  string `json:"resource"`
+	Used      int    `json:"used"`
+	Available int    `json:"available"`
```
falzm, line 14: Available can be inferred from `Max - Used`, feels redundant to me.

cmd/limits.go
```diff
@@ -0,0 +1,119 @@
...
+				Available: (int(limit.Max) - rUsed),
+				Used:      rUsed,
+			})
+		}
+	}
+
+	return &out, nil
+}
+
+func fetchUsedResources(rType string) (int, error) {
```
pierre-emmanuelJ, line 73: `rtype` can be `resourceType`, since you are using in the whole function

cmd/limits.go
```diff
@@ -0,0 +1,119 @@
...
+	var rUsed int
```
pierre-emmanuelJ, line 74: same here `rUsed` can be `resourceUsed`

cmd/limits.go
```diff
@@ -0,0 +1,117 @@
...
+type LimitsOutput []LimitsItemOutput
+
+func (o *LimitsOutput) toJSON()  { outputJSON(o) }
+func (o *LimitsOutput) toText()  { outputText(o) }
+func (o *LimitsOutput) toTable() { outputTable(o) }
+
+var limitsCmd = &cobra.Command{
+	Use:   "limits",
+	Short: "Current account limits",
+	Long: fmt.Sprintf(`This command lists the safety limits currently enforced on your account.
```
jeromer, line 26: s/enforced/attached/ ?

cmd/limits.go
```diff
@@ -0,0 +1,117 @@
...
+		"network":           "Private Networks",
+	}
+
+	out := LimitsOutput{}
+
+	for _, key := range limits {
+		limit := key.(*egoscale.ResourceLimit)
+
+		if display[limit.ResourceTypeName] != "" {
+			rUsed, err := fetchUsedResources(limit.ResourceTypeName)
```
jeromer, line 55: s/rUsed/used/

cmd/limits.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+
+	return &out, nil
+}
+
+func fetchUsedResources(resourceType string) (int, error) {
+	var resourceUsed int
+
+	switch resourceType {
```
jeromer, line 74: No default case in this `switch` ?

cmd/limits.go
```diff
@@ -0,0 +1,117 @@
...
+				Max:      int(limit.Max),
+				Used:     rUsed,
+			})
+		}
+	}
+
+	return &out, nil
+}
+
+func fetchUsedResources(resourceType string) (int, error) {
```
jeromer, line 71: (untested) 

```suggestion
func fetchUsedResources(resourceType string) (int, error) {
	switch resourceType {
	case "user_vm":
		instances, err := cs.ListWithContext(gContext, &egoscale.VirtualMachine{})
		if err != nil {
			return 0, err
		}

		return len(instances), nil

	case "snapshot":
		snapshots, err := cs.ListWithContext(gContext, &egoscale.Snapshot{})
		if err != nil {
			return 0, err
		}

		return len(snapshots), nil

	case "template":
		templates, err := cs.ListWithContext(gContext, &egoscale.Template{})
		if err != nil {
			return 0, err
		}

		return len(templates), nil

	case "network":
		networks, err := cs.ListWithContext(gContext, &egoscale.Network{})
		if err != nil {
			return 0, err
		}

		return len(networks), nil

	default: // public_elastic_ip
		eips, err := cs.ListWithContext(gContext, &egoscale.IPAddress{IsElastic: true})
		if err != nil {
			return 0, err
		}

		return len(eips), nil
	}
}

```

illi-j: No, because as you can see, there is more "resourceType" than we need:
[ResourceTypeName](https://github.com/exoscale/egoscale/blob/21b8548dd15850ce490130dc8f2aad57dcdc1a44/resource_limits.go#L8)

so in your case, if a resource is not in the display variable will be set to the `default` case

cmd/limits.go
illi-j: I referred to the description of the Exoscale console

cmd/limits.go
jeromer: ok.

cmd/limits.go
jeromer: ok

PR #233, opened 2020-04-23
Add go.mk support for exo cli

illi-j:


falzm:
```
$ make exo
Makefile:1: go.mk/init.mk: No such file or directory
make: *** No rule to make target 'go.mk/init.mk'.  Stop.
```

@illi-j is there something specific to do? If so, that'd be nice to either include it in the Makefile or document it in the README.

PR #237, opened 2020-04-24
Upgrade to Go 1.14

illi-j:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.13-stretch as builder
```
greut, line 1: you may even go with `1.14-buster`

PR #240, opened 2020-04-27
Full directory synchronization

janoszen:
This feature implements a full synchronization (upload and optional delete) for a local directory to the Exoscale Simple Object Storage.

Usage:

```
exo sos sync bucketname localdirectory remotepath
```

Optional arguments:

- `-r` or `--remove-deleted`: Remove locally deleted files from object storage.
- `-n` or `--dry-run`: Don't actually do anything, just print.
- `-c` or `--concurrency`: Sets the number of concurrent uploads (default 10).

QA: automated tests have been implemented for the diff algorithm, manual tests have been carried out for the full functionality.


cmd/sos_sync.go
```diff
@@ -0,0 +1,518 @@
...
+//Initializes the sync command
+func init() {
+	runE := newSosSyncRunE(sosSyncLiveClientFactory)
+	cmd := &cobra.Command{
+		Use:     "sync <bucket name> <local path> <remote-path>",
+		Short:   "Sync a local folder with the object storage",
+		Aliases: gSyncAlias,
+		RunE:    runE,
+	}
+	cmd.Flags().BoolP("remove-deleted", "r", false, "Remove remote files not present locally")
```
falzm, line 512: I suggest using the Rsync terminology, as many people are familiar with it:

```suggestion
	cmd.Flags().BoolP("remove-deleted", "d", false, "Delete remote files not present locally")
```

cmd/sos_sync.go
```diff
@@ -0,0 +1,518 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+//region Settings
```
falzm, line 20: Can you please remove extraneous comments? We try to use comments only when it's mandated by the linter or when it makes sense for the reader, to give context as why the code is implemented a specific way.

cmd/sos_sync.go
```diff
@@ -0,0 +1,518 @@
...
+
+const (
+	defaultParallelSosSync = 10
+)
+
+//endregion
+
+//region Type definitions
+
+type sosSyncClientFactory = func(certsFile string) (*sosClient, error)
```
falzm, line 30: It looks like you have somewhat of a Java background ;)
We strive to implement our Go project with the most idiomatic styles, namely:

* https://golang.org/doc/effective_go.html
* https://github.com/uber-go/guide/blob/master/style.md

For consistency's sake, could you please try to adopt a similar style to the rest of the codebase, so there is no additional cognitive load when operating across the entire project?

cmd/root.go
```diff
@@ -44,6 +44,7 @@ var gRevokeAlias = []string{"rvk"}
 var gShowAlias = []string{"get"}
 var gCreateAlias = []string{"add"}
 var gUploadAlias = []string{"up"}
+var gSyncAlias = []string{"s"}
```
pierre-emmanuelJ, line 47: IMHO, I don't think this alias is useful

janoszen:
> Thank you for this useful contribution @janoszen, however the proposed implementation feels way too over-engineered to me. Besides, we strive to implement our Go project with the most idiomatic styles, namely:
> 
> * https://golang.org/doc/effective_go.html
> * https://github.com/uber-go/guide/blob/master/style.md
> 
> For consistency's sake, could you please avoid using the factory design pattern and try to adopt a similar style to the rest of the codebase, so there is no additional cognitive load when operating across the entire project?

Thanks for the review.

I'm not entirely happy about the cognitive load on this code myself. However, I think having tests for something like sync, especially if we end up implementing more factors later, would be nice.

If we are not in a rush I can experiment with making the code more readable while also keeping the tests. Any pointers towards that goal would be welcome offline too.

falzm:
> I'm not entirely happy about the cognitive load on this code myself. However, I think having tests for something like sync, especially if we end up implementing more factors later, would be nice.

The tests are not the issue – quite the contrary to me, I'm happy to see more of them in this project that is lacking on this aspect. I was referring more to the Java-esque implementation, which clashes frankly with the global coding style and philosophy.
 
> If we are not in a rush I can experiment with making the code more readable while also keeping the tests. Any pointers towards that goal would be welcome offline too.

Definitely, let's bring this discussion internally and find a common ground. Thanks!

janoszen:
As discussed with @falzm I will attempt to refactor the code to be less convoluted but still keep the tests. If it succeeds we will merge it.

falzm:
As discussed with @janoszen, we'll pull the plug on this PR as the volume of code involved is overkill compared to other commands.

PR #244, opened 2020-05-06
Make certificates work on Windows again

janoszen:
This pull request implements checking the directory of the `exo` binary for a file named `sos-certs.pem` and load that if no `--certs-file` flag has been supplied.

This works around golang/go#16736 which prevents `exo sos` commands from working without explicitly passing a certificate chain.

~~Note that the build system still needs to be adapted to bundle `sos-certs.pem` into the ZIP file on Windows.~~

A separate PR #245 has been submitted as well which only warns the user.

### DIFF
--- a/cmd/sos.go
+++ b/cmd/sos.go
@@ -74,7 +74,7 @@ func newSOSClient(certsFile string) (*sosClient, error) {
 	if certsFile == "" && runtime.GOOS == "windows" {
 		// Check if the directory of the "exo" executable contains a file named "sos-certs.pem"
 		// to load the certificate chain from. This is done to work around Golang issue #16736
-		// on Windows ( https://github.com/golang/go/issues/16736 )
+		// on Windows (https://github.com/golang/go/issues/16736)
 		path, err := os.Executable()
 		if err == nil {
 			dir, err := filepath.Abs(filepath.Dir(path))
--- a/cmd/sos.go
+++ b/cmd/sos.go
@@ -30,6 +30,7 @@ var sosCmdLongHelp = func() string {
 
 	if runtime.GOOS == "windows" {
 		long += `
+
 IMPORTANT: Due to a bug in the Microsoft Windows support in the Go
 programming language (https://github.com/golang/go/issues/16736) Windows
 users are required to extract the sos-certs.pem file next to their exo.exe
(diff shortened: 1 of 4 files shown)


janoszen:
@falzm I added a bit of code to download the cert chain and bundle it for Windows, but I don't know if it will actually work as I can't test it on my setup.

falzm:
@janoszen I don't think it's safe to assume users will run `exo` straight from the decompressed archive directory, and that the certificate will always be besides the binary. At this point, if we really want to make it as transparent as possible for Windows users the best approach IMO is to embed the certificate content directly into the binary – which ironically was how the first implementation did: https://gist.github.com/falzm/4ba41b63ac794f84269377a0c45de862

janoszen:
@falzm happy to implement it that way too if needed, but two points:

1. Windows users are used to installers, not archives. This may be something to add in the future.
2. If they don't copy all the files but only the exe, that's a user fail. I can add an error message for that warning the user that the certificate chain file is missing and pointing them to the installation doc.

/cc @pyr

pyr:
@janoszen you may be more well versed in Windows idioms than we are. It feels a bit odd to me be but if we think this solves how people would use `exo` on windows for the common case let's do it this way.  We can hope golang gets to the bottom of their chain resolution issue at some point too.

falzm:
@janoszen I'm not familiar with Windows idioms either – although I'd have stored the certificate file besides the configuration file, personnally –, so if you're confident that going this way won't lead to support tickets or GitHub issues to complain about this then by all means.

janoszen:
@falzm I've added a better help text, other than that feel free to merge. At some point I will make an installer for it.

cmd/sos.go
```diff
@@ -56,6 +71,29 @@ func newSOSClient(certsFile string) (*sosClient, error) {
 		err error
 	)
 
+	if certsFile == "" && runtime.GOOS == "windows" {
+		// Check if the directory of the "exo" executable contains a file named "sos-certs.pem"
+		// to load the certificate chain from. This is done to work around Golang issue #16736
+		// on Windows ( https://github.com/golang/go/issues/16736 )
```
falzm, line 77:
```suggestion
		// on Windows (https://github.com/golang/go/issues/16736)
```

cmd/sos.go
```diff
@@ -56,6 +71,29 @@ func newSOSClient(certsFile string) (*sosClient, error) {
...
+				} else {
+					log.Warn(
+						"It seems you are running on Windows and your sos-certs.pem file is missing.\n" +
+							"Please download and extract all files from the exo cli release, not just the executable.\n" +
+							"Please see the 'exo sos help' command for more details.")
+				}
+			}
+		}
+	}
+
```
falzm, line 96: Ignoring errors is a dangerous bet that can lead to hours wasted on debugging obscure error messages that our users may report to us in the future (speaking from experience). A more typical way to implement this code is to check for errors early and handle them properly, which also grants a less indented code:

```go
	if err != nil {
		// Handle error
	}
	
	dir, err := filepath.Abs(filepath.Dir(path))
	if err != nil {
		// Handle error
	}

	tmpCertsFile := filepath.Join(dir, "sos-certs.pem")
	stat, err := os.Stat(tmpCertsFile)
	if err != nil {
		log.Warn(
			"It seems you are running on Windows and your sos-certs.pem file is missing.\n" +
			"Please download and extract all files from the exo cli release, not just the executable.\n" +
			"Please see the 'exo sos help' command for more details.")
		// Maybe return with error here instead of continuing
	}

	certsFile = tmpCertsFile
```

janoszen: OK, I overhauled that, but I'm unsure if I'm supposed to use log.Warn as it prints a JSON-formatted log message, which is not very user friendly. Any advice on that @falzm ?

falzm: You've imported the `log.Warn()` method from a new package `github.com/labstack/gommon`, which seems to log in JSON format by default. I'd prefer you use the standard `fmt.Fprintln(os.Stderr, "your message [...]")` call for this, please.

janoszen: Whoopsies, removed.

PR #247, opened 2020-05-07
Add support for experimental API calls

falzm:
This change brings a new `exoscale x` subcommands group code-generated
from the Exoscale API V2 OpenAPI specs using [openapi-cli-generator](https://github.com/danielgtaylor/openapi-cli-generator).

Example:

```console
$ ./bin/exo x -h
Low-level Exoscale API calls -- don't use this unless you have to.

These commands are automatically generated using openapi-cli-generator[1],
input parameters can be supplied either via stdin or using Shorthands[2].

[1]: https://github.com/danielgtaylor/openapi-cli-generator
[2]: https://github.com/danielgtaylor/openapi-cli-generator/tree/master/shorthand

Usage:
  exo x [command]

Available Commands:
  add-rule-to-security-group             add-rule-to-security-group
  add-service-to-load-balancer           add-service-to-load-balancer
  allocate-elastic-ip                    allocate-elastic-ip
  allocate-ipv6-on-instance              allocate-ipv6-on-instance
  attach-elastic-ip-on-instance          attach-elastic-ip-on-instance
  create-anti-affinity-group             create-anti-affinity-group
  create-instance                        create-instance
  create-instance-pool                   create-instance-pool
  create-load-balancer                   create-load-balancer
  create-security-group                  create-security-group
  delete-rule-from-security-group        delete-rule-from-security-group
  delete-service-from-load-balancer      delete-service-from-load-balancer
  delete-ssh-key                         delete-ssh-key
  destroy-anti-affinity-group            destroy-anti-affinity-group
  destroy-instance                       destroy-instance
  destroy-instance-pool                  destroy-instance-pool
  destroy-load-balancer                  destroy-load-balancer
  destroy-security-group                 destroy-security-group
  destroy-snapshot                       destroy-snapshot
  destroy-template                       destroy-template
  detach-elastic-ip-from-instance        detach-elastic-ip-from-instance
  elastic-ip                             elastic-ip
  generate-api-key                       generate-api-key
  generate-ssh-key                       generate-ssh-key
  get-api-key-details                    get-api-key-details
  get-result                             get-result
  help-config                            Show CLI configuration help
  help-input                             Show CLI input help
  import-ssh-key                         import-ssh-key
  index                                  index
  instance                               instance
  list-anti-affinity-group-instances     list-anti-affinity-group-instances
  list-anti-affinity-groups              list-anti-affinity-groups
  list-api-keys                          list-api-keys
  list-elastic-ips                       list-elastic-ips
  list-events                            list-events
  list-instance-pools                    list-instance-pools
  list-instance-types                    list-instance-types
  list-instances                         list-instances
  list-load-balancers                    list-load-balancers
  list-results                           list-results
  list-security-group-instances          list-security-group-instances
  list-security-groups                   list-security-groups
  list-snapshots                         list-snapshots
  list-ssh-keys                          list-ssh-keys
  list-templates                         list-templates
  list-zones                             list-zones
  ping                                   ping
  reboot-instance                        reboot-instance
  register-template                      register-template
  release-elastic-ip                     release-elastic-ip
  remove-elastic-ip-reverse-dns          remove-elastic-ip-reverse-dns
  remove-instance-reverse-dns            remove-instance-reverse-dns
  reset-instance-password                reset-instance-password
  restore-instance-from-snapshot         restore-instance-from-snapshot
  restore-instance-from-template         restore-instance-from-template
  revoke-api-key                         revoke-api-key
  show-instance-pool-details             show-instance-pool-details
  show-load-balancer-details             show-load-balancer-details
  show-load-balancer-service-details     show-load-balancer-service-details
  snapshot                               snapshot
  start-instance                         start-instance
  stop-instance                          stop-instance
  update-anti-affinity-group-description update-anti-affinity-group-description
  update-elastic-ip-healthcheck          update-elastic-ip-healthcheck
  update-elastic-ip-reverse-dns          update-elastic-ip-reverse-dns
  update-instance-description            update-instance-description
  update-instance-pool-description       update-instance-pool-description
  update-instance-pool-template          update-instance-pool-template
  update-instance-pool-topology          update-instance-pool-topology
  update-instance-pool-user-data         update-instance-pool-user-data
  update-instance-reverse-dns            update-instance-reverse-dns
  update-instance-security-groups        update-instance-security-groups
  update-instance-type                   update-instance-type
  update-instance-user-data              update-instance-user-data
  update-security-group-description      update-security-group-description
  update-security-group-rules            update-security-group-rules
  update-snapshot-description            update-snapshot-description
  version                                version

Flags:
  -h, --help            help for x
  -q, --query string    Filter / project results using JMESPath
      --raw             Output result of query as raw rather than an escaped JSON string or list
      --server string   Override server URL
      --verbose         Enable verbose log output

Global Flags:
  -Q, --quiet                    Quiet mode (disable non-essential command output)

Use "exo x [command] --help" for more information about a command.
```


falzm:
Note: I had to copy egoscale's complete implementation of the API V2 request signature mechanism, as the code generated by `openapi-cli-generator` doesn't use egoscale at all, and also that the code is actually located in an internal package which is not usable by external modules. This is a trade-off that I'm personally fine with, but let me know if you have better options to suggest.

pyr:
This is really great stuff and will simplify the transition over to our new API. A kind request would be to do three things on top of this:

- Infer the target zone from the `defaultZone` configuration param
- Allow overriding the zone with a `-Z` parameter
- Infer the API endpoint from the one present in config (interpolating the zone and replacing the suffix).

With this, we could simply `exo x ...` by default and `exo x -Z de-fra-1 ...`, the `--server` option can still prove useful and it should override this but for the common case it would be very convenient.

pierre-emmanuelJ:
> Note: I had to copy egoscale's complete implementation of the [API V2 request signature mechanism](https://github.com/exoscale/egoscale/blob/master/internal/v2/security.go), as the code generated by `openapi-cli-generator` doesn't use egoscale at all, and also that the code is actually located in an internal package which is not usable by external modules. This is a trade-off that I'm personally fine with, but let me know if you have better options to suggest.

Maybe we can move v2 from internal to a public package, or the idea can be to use git branch for package versioning. What do you think about that?

.travis.yml
```diff
@@ -8,22 +8,19 @@ go:
 cache: apt
 
 env:
-  - GO111MODULE=on HUGO_VERSION=0.58.3 GOLANGCI_LINT_VERSION=1.21.0
```
pyr, line 11: You could delte the whole file

cmd/x.go
```diff
@@ -0,0 +1,67 @@
...
+	xCmd.PersistentPreRunE = func(cmd *cobra.Command, _ []string) error {
+		// If no value is provided for flag `--server`, infer the server URL from the current
+		// profile's default zone or explicit `--zone` flag if specified:
+		if server, _ := cmd.Flags().GetString("server"); server == "" {
+			zone := gCurrentAccount.DefaultZone
+			if z, _ := cmd.Flags().GetString("zone"); z != "" {
+				zone = z
+			}
+
+			env, err := cmd.Flags().GetString("environment")
```
pyr, line 41: I still think we should allow setting the environment in our profiles (the following would assume there's an `Env` field in accounts:

```suggestion
			env, _ := cmd.Flags().GetString("environment"); env == "" {
			  if (gCurrentAccount.Env != "") {
			      env = gCurrentAccount.Env
			  } else {
			    env = "api"
			  }
			}
```

falzm: Done.

PR #249, opened 2020-05-12
vm: snapshot export: add flag --download

illi-j:
```
exo vm snapshot export f5860e58-fa71-48b1-9724-4b4c3a3759b5 --download ./
Exporting snapshot "f5860e58-fa71-48b1-9724-4b4c3a3759b5". success
Downloading snapshot file. 6.20 MiB / 452.12 MiB [>-------------------------------------------------------------| 3m16s ]

ls *.qcow2
f5860e58-fa71-48b1-9724-4b4c3a3759b5.qcow2
```

```
exo vm snapshot export f5860e58-fa71-48b1-9724-4b4c3a3759b5 --download template.qcow2
Exporting snapshot "f5860e58-fa71-48b1-9724-4b4c3a3759b5". success
Downloading snapshot file. success
Verifying downloaded file checksum. success

ls *.qcow2
template.qcow2
```

### DIFF
--- a/cmd/snapshot_export.go
+++ b/cmd/snapshot_export.go
@@ -1,11 +1,22 @@
 package cmd
 
 import (
+	"crypto/md5"
+	"encoding/hex"
+	"errors"
 	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/exoscale/egoscale"
 	"github.com/spf13/cobra"
+	"github.com/vbauerster/mpb/v4"
+	"github.com/vbauerster/mpb/v4/decor"
 )
 
 type snapshotExportOutput struct {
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -25,7 +25,8 @@ require (
 	github.com/spf13/pflag v1.0.3
 	github.com/spf13/viper v1.3.1
 	github.com/vbauerster/mpb/v4 v4.8.4
-	golang.org/x/crypto v0.0.0-20191112222119-e1110fd1c708
+	github.com/vbauerster/mpb/v5 v5.0.4
```
falzm, line 28: Is it *really* necessary to have both `vbauerster/mpb` v4 and v5? IMO you should either use v4 for this command, or upgrade the rest of the code to use v5 too.

illi-j: https://github.com/exoscale/cli/pull/249/commits/78d3bdd90981fda1c318c1818c526bfdb2a15da1

cmd/snapshot_export.go
```diff
@@ -17,38 +28,172 @@ func (o *snapshotExportOutput) toJSON()  { outputJSON(o) }
...
+		if !cmd.Flags().Changed("download") {
+			return output(&snapshotExportOutput{
+				URL:      snapshot.PresignedURL,
+				Checksum: snapshot.MD5sum,
+			}, nil)
+		}
+
+		if err = downloadExportedSnapshot(filePath, args[0], snapshot.PresignedURL, snapshot.MD5sum); err != nil {
+			return err
+		}
```
falzm, line 62: `downloadExportedSnapshot` should do only what its name implies: download the template:

```suggestion
		if err = downloadExportedSnapshot(filePath, args[0], snapshot.PresignedURL); err != nil {
			return err
		}
```

Then, if no error occurred during template download you can call `checkExportedSnapshot`:

```go
		if err = checkExportedSnapshot(filePath, snapshot.MD5sum); err != nil {
			return err
		}
```

cmd/snapshot_export.go
```diff
@@ -17,38 +28,172 @@ func (o *snapshotExportOutput) toJSON()  { outputJSON(o) }
...
-	snapshot := res.(*egoscale.ExportSnapshotResponse)
+	return res.(*egoscale.ExportSnapshotResponse), nil
+}
+
+func downloadExportedSnapshot(filePath, snapshotID, url string) (string, error) {
+	filePath = filepath.ToSlash(filePath)
+
+	if filePath[len(filePath)-1:] == "/" {
+		filePath = filePath + snapshotID
+	}
```
falzm, line 92: If I understand correctly what you do here (i.e. checking if the specified path is a directory, and if so append the snapshot ID to it as filename), it's a rather fragile way to do so: a path can point a directory without being terminated with a trailing slash:

```console
$ mkdir blah
$ stat -c %F blah
directory
$ stat -c %F .
directory
```

Instead, you should call [`os.Stat(filePath)`](https://pkg.go.dev/os?tab=doc#Stat), and if it doesn't return an error – which you'll check using [`os.IsNotExist(err)`](https://pkg.go.dev/os?tab=doc#IsExist) –  check what the returned os.FileInfo's `IsDir()` method tells you instead.

cmd/snapshot_export.go
```diff
@@ -17,38 +28,172 @@ func (o *snapshotExportOutput) toJSON()  { outputJSON(o) }
...
+	hash := md5.New()
+
+	if _, err := io.Copy(hash, file); err != nil {
+		return err
+	}
+
+	h := hex.EncodeToString(hash.Sum(nil))
+
+	if h != md5sum {
+		return errors.New("checksum mismatch")
```
falzm, line 188: It might be helpful to the user to actually return the checksums:

```suggestion
		return fmt.Errorf("checksum mismatch: expected %q, got %q", md5sum, h)
```

cmd/snapshot_export.go
```diff
@@ -17,38 +28,172 @@ func (o *snapshotExportOutput) toJSON()  { outputJSON(o) }
...
+
+	if err = file.Close(); err != nil {
+		return "", err
+	}
+
+	return localFilePath, nil
+}
+
+func checkExportedSnapshot(filePath, md5sum string) error {
+	fmt.Print("Verifying downloaded file checksum... ")
```
falzm, line 170: As mentioned earlier, no messages should be output outside of the main CLI command function: this should be moved at caller level.

cmd/snapshot_export.go
falzm: Besides, you should return an error telling the user that directory paths are not allowed instead of silently appending the snapshot ID to the directory path.

cmd/snapshot_export.go
```diff
@@ -17,38 +27,175 @@ func (o *snapshotExportOutput) toJSON()  { outputJSON(o) }
...
+
+	st, err := os.Stat(filePath)
+	if err == nil {
+		if st.IsDir() {
+			filePath = filepath.Join(filePath, snapshotID+".qcow2")
+			_, err = os.Stat(filePath)
+			if !os.IsNotExist(err) {
+				return "", fmt.Errorf("file %q: already exists", filePath)
+			}
+		}
```
pierre-emmanuelJ, line 101: Here if the path is not a dir, I think you can return the same error as above 
```Golang
return "", fmt.Errorf("file %q: already exists", filePath)
```

cmd/snapshot_export.go
```diff
@@ -17,38 +27,177 @@ func (o *snapshotExportOutput) toJSON()  { outputJSON(o) }
...
+		} else {
+			return "", fmt.Errorf("file %q: already exists", filePath)
+		}
+	}
 
-	return &snapshotExportOutput{
-		URL:      snapshot.PresignedURL,
-		Checksum: snapshot.MD5sum,
-	}, nil
+	if err != nil {
```
pierre-emmanuelJ, line 106: Here if `err` from `st, err := os.Stat(filePath)`  is not nil you should check before if it's not a standard error.
like here: https://github.com/exoscale/cli/blob/master/cmd/sos_download.go#L63-L68

cmd/snapshot_export.go
```diff
@@ -17,38 +27,177 @@ func (o *snapshotExportOutput) toJSON()  { outputJSON(o) }
...
+		if filepath.Ext(filePath) != ".qcow2" {
+			filePath = filePath + ".qcow2"
+		}
+		_, err = os.Stat(filePath)
```
pierre-emmanuelJ, line 110: here same as above, `err` can be checked

PR #251, opened 2020-05-22
feat(sos-download): Add flag that allows overwriting the destination

wdullaer:
Adds a flag to `exo sos download` which will skip the file existence check.
It's off by default.

There are scenarios where this is valid behaviour.

Slightly related: I think that the `os.Rename` step at https://github.com/exoscale/cli/compare/sos-download-overwrite?expand=1#diff-cbe63d0320814a34674804f20dcfac94R140 does not take into account the case where `localFilePath` is a directory, and thus needs to be appended with `objectName`.

### DIFF
--- a/cmd/sos_download.go
+++ b/cmd/sos_download.go
@@ -46,24 +46,18 @@ var downloadCmd = &cobra.Command{
 		}
 
 		// Verify if destination already exists.
-		st, err := os.Stat(localFilePath)
-		if err == nil {
-			// If the destination exists and is a directory.
-			if st.IsDir() {
-				localFilePath = path.Join(localFilePath, objectName)
-				_, err = os.Stat(localFilePath)
-				if err == nil {
-					return fmt.Errorf("file %q: already exists", localFilePath)
-				}
-			} else {
-				return fmt.Errorf("file %q: already exists", localFilePath)
-			}
+		force, err := cmd.Flags().GetBool("force")
+		if err != nil {
+			return err
 		}
 
-		// Proceed if file does not exist. return for all other errors.
-		if err != nil {
-			if !os.IsNotExist(err) {
+		if !force {
+			exists, err := destinationExists(localFilePath, objectName)
+			if err != nil {
 				return err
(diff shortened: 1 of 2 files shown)


cmd/sos_download.go
```diff
@@ -159,6 +148,32 @@ var downloadCmd = &cobra.Command{
...
+		if !os.IsNotExist(err) {
+			return false, err
+		}
+	}
+	return false, nil
+}
+
 func init() {
 	sosCmd.AddCommand(downloadCmd)
+	downloadCmd.Flags().Bool("allow-overwrite", false, "Allow overwriting the destination")
```
falzm, line 178: I'd prefer we go with the usual *parlance* in CLI tools:

```suggestion
	downloadCmd.Flags().BoolP("force", "f", false, "Overwrite the destination file if it already exists")
```

cmd/sos_download.go
```diff
@@ -46,24 +46,13 @@ var downloadCmd = &cobra.Command{
...
-			if !os.IsNotExist(err) {
+		if allowOverwrite, err := cmd.Flags().GetBool("allow-overwrite"); err != nil {
+			return err
+		} else if !allowOverwrite {
+			if exists, err := destinationExists(localFilePath, objectName); err != nil {
 				return err
+			} else if exists {
+				return fmt.Errorf("file %q: already exists", localFilePath)
 			}
 		}
```
pierre-emmanuelJ, line 57: An alternative implementation
```Golang
allowOverwrite, err := cmd.Flags().GetBool("allow-overwrite")
if err != nil {
	return err
}

if !allowOverwrite {
	exists, err := destinationExists(localFilePath, objectName)
    if err != nil {
		return err
	}

    if exists {
		return fmt.Errorf("file %q: already exists", localFilePath)
	}
}
```

cmd/sos_download.go
wdullaer: Will do. I'll refactor the variable names to `force` as well.

Issue #181, opened 2019-09-10
Push new versions to Docker hub


mcanevet:
Latest version on Docker Hub is 1.3.0. Could you please push new versions?

falzm:
Thank you for reporting this problem, sorry about that... Images for the latest versions are now available 👍

PR #254, opened 2020-06-05
Use Alpine base

J0WI:
> Alpine Linux is a Linux distribution built around musl libc and BusyBox. The image is only 5 MB in size and has access to a package repository that is much more complete than other BusyBox based images. This makes Alpine Linux a great image base for utilities and even production applications.

https://hub.docker.com/_/alpine

This simplifies things a lot.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.14-buster as builder
+FROM golang:1.14-alpine as builder
 
 ADD . /src
 WORKDIR /src
@@ -10,7 +10,7 @@ ENV CGO_ENABLED=1
 RUN go build -mod vendor -o exo \
         -ldflags "-s -w -X main.version=${VERSION} -X main.commit=${VCS_REF}"
 
-FROM ubuntu:eoan
+FROM alpine:3.12
 
 ARG VERSION
 ARG VCS_REF
@@ -26,14 +26,7 @@ LABEL org.label-schema.build-date=${BUILD_DATE} \
       org.label-schema.url="https://exoscale.github.io/cli" \
       org.label-schema.schema-version="1.0"
 
-RUN set -xe \
- && apt-get update -q \
- && apt-get upgrade -q -y \
- && apt-get install -q -y \
-        ca-certificates \
- && apt-get autoremove -y \
- && apt-get clean \
- && rm -rf /var/lib/apt/lists/*
+RUN apk add --no-cache ca-certificates
 
 COPY --from=builder /src/exo /
 ENTRYPOINT ["/exo"]


greut:
If eoan should read focal, don't you have any issues regarding CGO when running it?

J0WI:
> If eoan should read focal

Not sure what you mean here. The patch replaces Ubuntu completely.

pierre-emmanuelJ:
Hello @J0WI 

Thank you for your patch and your time! 😃 

I have just tested to build the image and put it in [scratch](https://hub.docker.com/_/scratch) instead of `ubuntu:eoan` or `alpine`.

This is my suggestion:
https://github.com/exoscale/cli/commit/539593601e0a26070732a6e9367919a1c9e30336

scratch seems to be a good way to reduce image size and contain a single binary.

greut:
@J0WI looks ok!

@pierre-emmanuelJ if the image is smaller, you still need the certificates. E.g. with [linuxkit/ca-certificates](https://hub.docker.com/r/linuxkit/ca-certificates)

Then for CI usage, having more than nothing is sometimes useful.

pierre-emmanuelJ:
You are right, I forgot the CA certs, this image seems good https://hub.docker.com/r/linuxkit/ca-certificates.

https://github.com/linuxkit/linuxkit/blob/master/pkg/ca-certificates/Dockerfile 👍

greut:
@pierre-emmanuelJ for CI usage, you need a shell. E.g. [nancy](https://github.com/sonatype-nexus-community/nancy) (great soft btw), produces two images, a bare one and an alpine one for CI usage.

https://github.com/sonatype-nexus-community/nancy/blob/master/.goreleaser.yml

pierre-emmanuelJ:
As discussed with @greut I updated my suggestion:
https://github.com/exoscale/cli/commit/521318cfe882f3854b24bfd115da3dc374538a0c

tested:
```
docker run -v /home/userx/.exoscale:/root/ exoscale/cli:latest zone
current user cannot be read, using "root"
|                  ID                  |   NAME   |
|--------------------------------------|----------|
| b0fcd72f-47ad-4779-a64f-fe4de007ec72 | at-vie-1 |
| a4f2d4d1-0583-425e-9945-e6d891524093 | bg-sof-1 |
| 381d0a95-ed4a-4ad9-b41c-b97073c1a433 | ch-dk-2  |
| 1747ef5e-5451-41fd-9f1a-58913bae9702 | ch-gva-2 |
| de88c980-78f6-467c-a431-71bcc88e437f | de-fra-1 |
| 84e0eb97-bf03-4ed7-999f-96897f08778b | de-muc-1 |
```

pierre-emmanuelJ:
> @pierre-emmanuelJ for CI usage, you need a shell. E.g. [nancy](https://github.com/sonatype-nexus-community/nancy) (great soft btw), produces two images, a bare one and an alpine one for CI usage.
> 
> https://github.com/sonatype-nexus-community/nancy/blob/master/.goreleaser.yml

Yes could be doable! To add another Dockerfile for the CI for testing..etc.
And have a little and clean image with scratch for the official image.

J0WI:
> As discussed with @greut I updated my suggestion:
> [521318c](https://github.com/exoscale/cli/commit/521318cfe882f3854b24bfd115da3dc374538a0c)

Please do not depend on unofficial base images.

How about coping `ca-certificates` from the builder stage? They are already installed there.
You can also use `FROM busybox` as a base image if you don't need the ability to install any dependency.

pierre-emmanuelJ:
> Please do not depend on unofficial base images.

LinuxKit is not a little organization by the way 😉 

> How about coping `ca-certificates` from the builder stage? They are already installed there.
> You can also use `FROM busybox` as a base image if you don't need the ability to install any dependency.

This is a good example:
https://github.com/linuxkit/linuxkit/blob/master/pkg/ca-certificates/Dockerfile

J0WI:
> LinuxKit is not a little organization by the way

That's worth nothing if their base images have not the same quality as upstream

falzm:
I have a naive question: what use case prompts for a dedicated Docker image of the `exo` CLI? I've personally never used it in the context of a Docker/k8s setup, but if I had to I'd rather install the CLI directly by downloading one of the [pre-built releases](https://github.com/exoscale/cli/releases) directly using `wget`/`curl` in a purpose-built Docker image if I need it.

falzm:
If nobody has a valid use case justifying that we should keep maintaining a Docker image of a single binary command, I suggest we stop doing it and let people that have an opinion about which base image to use do it themselves ¯\\\_(ツ)_/¯

greut:
@mcanevet ? https://github.com/exoscale/cli/issues/181

mcanevet:
Please continue to provide an official image. We are using it in our Gitlab CI pipeline.

pierre-emmanuelJ:
We closed this PR since this one https://github.com/exoscale/cli/pull/262 is merged.
Thanks @J0WI for your patch.

pierre-emmanuelJ:
Little update with Alpine: https://github.com/exoscale/cli/pull/263

PR #255, opened 2020-06-05
vm: new "update" command

falzm:
This change introduces a new `exo vm update` command enabling users to
edit some Compute instance properties, such as the display/hostname and
the cloud-init user data.

Besides, a new `exo vm show --user-data` flag now returns the current cloud-init user data configuration of a Compute instance.

### DIFF
--- a/cmd/vm_edit.go
+++ b/cmd/vm_edit.go
@@ -57,9 +57,11 @@ var vmEditCmd = &cobra.Command{
 			if !gQuiet {
 				fmt.Println("Virtual machine updated successfully")
 			}
+
+			return nil
 		}
 
-		return nil
+		return cmd.Usage()
 	},
 }
 
--- a/cmd/vm_edit.go
+++ b/cmd/vm_edit.go
@@ -1,0 +1,70 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/exoscale/egoscale"
+	"github.com/spf13/cobra"
+)
+
+var vmEditCmd = &cobra.Command{
+	Use:   "edit <name|ID>",
+	Short: "Edit virtual machine properties",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		var (
+			vmEdit egoscale.UpdateVirtualMachine
+			edited bool
+		)
+
+		if len(args) != 1 {
+			return cmd.Usage()
+		}
+
+		vm, err := getVirtualMachineByNameOrID(args[0])
+		if err != nil {
+			return err
+		}
+		vmEdit.ID = vm.ID
+
+		name, err := cmd.Flags().GetString("name")
+		if err != nil {
+			return err
+		}
+		if cmd.Flags().Changed("name") {
+			vmEdit.Name = name
+			vmEdit.DisplayName = name
+			edited = true
+		}
+
+		userDataPath, err := cmd.Flags().GetString("cloud-init-file")
+		if err != nil {
+			return err
+		}
+		if userDataPath != "" {
+			vmEdit.UserData, err = getUserDataFromFile(userDataPath)
+			if err != nil {
+				return err
+			}
+			edited = true
+		}
+
+		if edited {
+			_, err = cs.RequestWithContext(gContext, &vmEdit)
+			if err != nil {
+				return fmt.Errorf("unable to update virtual machine: %s", err)
+			}
+
+			if !gQuiet {
+				fmt.Println("Virtual machine updated successfully")
+			}
+		}
+
+		return nil
+	},
+}
+
+func init() {
+	vmEditCmd.Flags().String("name", "", "display name")
+	vmEditCmd.Flags().String("cloud-init-file", "", "path to a cloud-init user data file")
+	vmCmd.AddCommand(vmEditCmd)
+}
--- a/cmd/instance_pool_create.go
+++ b/cmd/instance_pool_create.go
@@ -113,13 +113,9 @@ Supported output template annotations: %s`,
 		userData := ""
 
 		if userDataPath != "" {
-			userData, err = getUserData(userDataPath)
+			userData, err = getUserDataFromFile(userDataPath)
 			if err != nil {
 				return err
-			}
-
-			if len(userData) >= maxUserDataLength {
-				return fmt.Errorf("user-data maximum allowed length is %d bytes", maxUserDataLength)
 			}
 		}
 
--- a/cmd/instance_pool_update.go
+++ b/cmd/instance_pool_update.go
@@ -71,13 +71,9 @@ Supported output template annotations: %s`,
 
 		userData := ""
 		if userDataPath != "" {
-			userData, err = getUserData(userDataPath)
+			userData, err = getUserDataFromFile(userDataPath)
 			if err != nil {
 				return err
-			}
-
-			if len(userData) >= maxUserDataLength {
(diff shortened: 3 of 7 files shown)


falzm:
[ch14907]

cmd/vm_edit.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+		if userDataPath != "" {
+			vmEdit.UserData, err = getUserDataFromFile(userDataPath)
+			if err != nil {
+				return err
+			}
+			edited = true
+		}
+
+		if edited {
```
pierre-emmanuelJ, line 51: IMO I Don't think the `edited` bool is useful, I've tried with cs you can update an instance without parameters, only with ID it's works and update nothing.

falzm: @pierre-emmanuelJ good to know, but in this case it also allows us to give a feedback to the user, signaling that the command *actually* changed something. Otherwise, we'd always return "Virtual machine updated successfully" even though we didn't do anything, which is confusing from a UX standpoint.

ste-m: I would also be in favour of not returning "Virtual machine updated successfully" if no flag is present.

I can as well observe the following behaviour in other commands:

    $ exo vm scale
    error: required flag(s) "service-offering" not set

    $ exo vm updateip
    Usage:
      exo vm updateip <vm name|id> <network name|id> [flags]
      ...
      
So i'd say that imho the best would be to have something like

    $ exo vm edit
    Usage:
      exo vm edit <vm name|id> [flags]
      ...

for non conform calls of the command.

falzm: @ste-m fixed as suggested:

```console
$ ./bin/exo vm edit testvm
Usage:
  exo vm edit <name|ID> [flags]

Flags:
      --cloud-init-file string   path to a cloud-init user data file
  -h, --help                     help for edit
      --name string              display name

Global Flags:
  -Q, --quiet                    Quiet mode (disable non-essential command output)
```

exo-cedric:
I'm wondering whether `edit` is a good choice for the command name; it makes me think that an _editor_ might be involved, which would allow me to see the previous setting and actually _edit_ it; unless I misread the code, it seems this is not the case (please correct me if I'm wrong).

And this raises the question of how to recover (dump) existing _user-data_ (again, I may have missed the ad-hoc command).

I would suggest:
- add `exo vm show --user-data`
- a generic `exo vm update [--ip <ip-address>] [--name <display-name>] [--user-data <user-data-file>] [whatever else we may want to update in the future]`

falzm:
> I'm wondering whether `edit` is a good choice for the command name; it makes me think that an _editor_ might be involved, which would allow me to see the previous setting and actually _edit_ it; unless I misread the code, it seems this is not the case (please correct me if I'm wrong).

@exo-cedric I went for *edit* because unfortunately there is already a `updateip` command, and having both looks sloppy to me UX-wise. Had I had the opportunity I'd have named it `update`, I agree that it's a better suited term and more consistent with a similar operation in other commands

> And this raises the question of how to recover (dump) existing _user-data_ (again, I may have missed the ad-hoc command).
> 
> I would suggest:
> 
>     * add `exo vm show --user-data`
> 
>     * a generic `exo vm update [--ip <ip-address>] [--name <display-name>] [--user-data <user-data-file>] [whatever else we may want to update in the future]`

The user data of an instance is not returned along with its other properties, a dedicated API call `GetVirtualMachineUserData` has to be performed to do that. I can implement it if it's deemed useful, but since even the portal doesn't display existing user-data I've skipped retrieving it it as well. WDYT?

exo-cedric:
> unfortunately there is already a updateip command, and having both looks sloppy to me UX-wise. Had I had the opportunity I'd have named it update

Could we imagine going for `update` and deprecating (yet keeping, for the time being) `updateip` ?
(this is a naive question; if it goes against existing guidelines, forget about it ;-) )

> I can implement it if it's deemed useful, but since even the portal doesn't display existing user-data I've skipped retrieving it it as well. WDYT?

Maybe its not being implemented in the portal is due to security considerations ?

On the other hand, retrieving _existing_ user-data before their update sounds almost like a _must_ to me (you would not want to remove key parameters from what is already configured).

Afterthoughts: one might argue that properly authorized users can retrieve these data directly from within their VM, using the `http://169.254.169.254/1.0/user-data` endpoint or looking in `/var/lib/cloud/instance/user-data.txt.i`; I guess it comes down to what level of user-friendliness we wish to achieve

falzm:
> > unfortunately there is already a updateip command, and having both looks sloppy to me UX-wise. Had I had the opportunity I'd have named it update
> 
> Could we imagine going for `update` and deprecating (yet keeping, for the time being) `updateip` ?
> (this is a naive question; if it goes against existing guidelines, forget about it ;-) )
> 
> > I can implement it if it's deemed useful, but since even the portal doesn't display existing user-data I've skipped retrieving it it as well. WDYT?

You're right, `update` makes more sense overall – I'll rename the command. I have no idea if the `updateip` command is actually useful, but for the moment i'll have to stay just in case someone uses it...

> Maybe its not being implemented in the portal is due to security considerations ?

Maybe @ste-m has an explanation?

> On the other hand, retrieving _existing_ user-data before their update sounds almost like a _must_ to me (you would not want to remove key parameters from what is already configured).
> 
> Afterthoughts: one might argue that properly authorized users can retrieve these data directly from within their VM, using the `http://169.254.169.254/1.0/user-data` endpoint or looking in `/var/lib/cloud/instance/user-data.txt.i`; I guess it comes down to what level of user-friendliness we wish to achieve

The problem is that the content of the user data can be quite large, so UX-wise it's not possible to display them properly along with the other key/value properties of a VM as done in `exo vm show`. I'm thinking about a `exo vm show --user-data` extension flag that would retrieve and display only those to the user, would that work for you?

exo-cedric:
> The problem is that the content of the user data can be quite large, so UX-wise it's not possible to display them properly along with the other key/value properties of a VM as done in `exo vm show`. I'm thinking about a `exo vm show --user-data` extension flag that would retrieve and display only those to the user, would that work for you?

Actually, it would be ideal (especially if no further formatting/whatever took place), allowing for the following workflow:
```bash
exo vm show --user-data ${VM_UUID} > /tmp/user-data
${EDITOR} /tmp/user-data
exo vm update --user-data /tmp/user-data ${VM_UUID}
```

falzm:
> > The problem is that the content of the user data can be quite large, so UX-wise it's not possible to display them properly along with the other key/value properties of a VM as done in `exo vm show`. I'm thinking about a `exo vm show --user-data` extension flag that would retrieve and display only those to the user, would that work for you?
> 
> Actually, it would be ideal (especially if no further formatting/whatever took place), allowing for the following workflow:
> 
> ```shell
> exo vm show --user-data ${VM_UUID} > /tmp/user-data
> ${EDITOR} /tmp/user-data
> exo vm update --user-data /tmp/user-data ${VM_UUID}
> ```

We have a deal, then :)

falzm:
@exo-cedric PR updated as discussed.

cmd/vm_edit.go
pierre-emmanuelJ: @ste-m  as discussed earlier with @falzm in PM, would be great to return the `vm show` output  as result to see the changes, if some flags is set of course.

Since the call `getVirtualMachineUserData` is in the API maybe we can do something with that to add it in the output result. I think it can be also useful to support the discussion below with @exo-cedric and @falzm.

PR #256, opened 2020-06-10
Add support for Network Load Balancer resources management

falzm:
This change implements support for Network Load Balancer resources
management.


.golangci.yml
```diff
@@ -4,7 +4,6 @@ linters-settings:
 
 linters:
   enable:
-    - goconst
```
pierre-emmanuelJ, line 7: not possible with `// nolint:goconst` ?

falzm: I don't believe this check is relevant, to be honest.

cmd/nlb_list.go
```diff
@@ -0,0 +1,91 @@
...
+
+	out := make(nlbListOutput, 0)
+	res := make(chan nlbListItemOutput)
+	defer close(res)
+
+	go func() {
+		for nlb := range res {
+			out = append(out, nlb)
+		}
+	}()
```
pierre-emmanuelJ, line 61: IMO this go routine is not useful in this case, can be appended directly in the result loop.

cmd/nlb_service_show.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func (o *nlbServiceShowOutput) toJSON() { outputJSON(o) }
+func (o *nlbServiceShowOutput) toText() { outputText(o) }
+func (o *nlbServiceShowOutput) toTable() {
+	t := table.NewTable(os.Stdout)
+	defer t.Render()
+
+	t.SetHeader([]string{"NLB Service"})
+	t.Append([]string{"ID", o.ID})
```
pierre-emmanuelJ, line 52: `output()` dedicated function can not be used in this case?

cmd/nlb_list.go
falzm: Tempting, but after a quick search it looks like it's not a good idea to append to a slice from multiple goroutines:

https://groups.google.com/g/golang-nuts/c/SRtSTar853g?pli=1
https://medium.com/@cep21/gos-append-is-not-always-thread-safe-a3034db7975
http://dnaeon.github.io/concurrent-maps-and-slices-in-go/

Alternatively, we could wrap the slice in a new type that includes a `sync.Mutex` for locking, but isn't it overkill?

pierre-emmanuelJ: After looking more deeply, didn't see your goroutine in `forEachZone `. Well let's keep like that 😉

pierre-emmanuelJ: >Alternatively, we could wrap the slice in a new type that includes a sync.Mutex for locking, but isn't it overkill?

Yes let's keep the go routine with the chan in this case

PR #262, opened 2020-06-22
Dockerfile: Make Docker image smaller

pierre-emmanuelJ:
This PR referring to https://github.com/exoscale/cli/pull/254, to make the docker image size smaller.
Fix #254


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.14.4-alpine as builder
+FROM golang:1.14-alpine as builder
```
illi-j, line 1: 👍

PR #263, opened 2020-06-23
refactor: Use alpine as cli base image

wdullaer:
The alpine base image still provides a shell in the container. This
makes it much easier to use the image in a CI environment or as an init
container in a k8s pod.
The added utility should be worth the extra 6Mb in image size.

```
REPOSITORY                         TAG                 IMAGE ID            CREATED             SIZE
cli                                test                d9c1593fd3bf        5 minutes ago       26.7MB
exoscale/cli                       latest              ec194dd39256        19 hours ago        20.8MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,14 +10,9 @@ ENV CGO_ENABLED=0
 RUN go build -a -mod vendor -o exo \
         -ldflags "-s -w -X main.version=${VERSION} -X main.commit=${VCS_REF}"
 
-FROM alpine:3.12.0 as ca-certificates
-
-RUN apk add --no-cache ca-certificates
-
-FROM scratch
+FROM alpine:3.12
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -10,14 +10,9 @@ ENV CGO_ENABLED=0
 RUN go build -a -mod vendor -o exo \
         -ldflags "-s -w -X main.version=${VERSION} -X main.commit=${VCS_REF}"
 
-FROM alpine:3.12.0 as ca-certificates
-
-RUN apk add --no-cache ca-certificates
-
-FROM scratch
+FROM alpine:3.12.0
```
pierre-emmanuelJ, line 13: Is it possible to replace with `3.12` to automatically pull the latest bugfix of `3.12` thanks :)

PR #265, opened 2020-06-24
nlb: support referencing resources by name or ID

falzm:
This change improves the `exo nlb` commands by allowing users to
reference NLB instances (and services) by their name or ID.

### DIFF
--- a/cmd/nlb.go
+++ b/cmd/nlb.go
@@ -1,6 +1,12 @@
 package cmd
 
-import "github.com/spf13/cobra"
+import (
+	"context"
+	"fmt"
+
+	"github.com/exoscale/egoscale"
+	"github.com/spf13/cobra"
+)
 
 var nlbCmd = &cobra.Command{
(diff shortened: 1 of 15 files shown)


falzm:
⚠️ Don't merge this PR until we confirm that the API enforces resources names uniqueness.

falzm:
[ch15681]

cmd/nlb.go
```diff
@@ -10,3 +16,19 @@ var nlbCmd = &cobra.Command{
 func init() {
 	RootCmd.AddCommand(nlbCmd)
 }
+
+// lookupNLB attempts to look up an NLB resource by name or ID.
+func lookupNLB(ctx context.Context, zone, ref string) (*egoscale.NetworkLoadBalancer, error) {
```
pierre-emmanuelJ, line 21: Is it possible to get directly in the API by name?

falzm: No, ID only.

PR #267, opened 2020-07-06
Add "scp" command

falzm:
This change introduces a new `exo scp` command, similar to `exo ssh`
but for sending/retrieve files to/from a Compute instance.


falzm:
[ch16135]

PR #273, opened 2020-07-28
Add dynamic names autocompletion for VM-related commands

falzm:
This change introduces dynamic CLI argument auto-completion for `exo vm *`
commands (and `exo ssh`) with the current user's Compute instance names:

[asciicast](https://asciinema.org/a/7IcN1UacbwtHoZFLr8Fi4TnYK)

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -57,3 +57,13 @@ func cmdExitOnUsageError(cmd *cobra.Command, reason string) {
 	cmd.Usage() // nolint:errcheck
 	os.Exit(1)
 }
+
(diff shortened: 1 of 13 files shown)


falzm:
> 👍
> (wondering however how this will behave with exoscale-ops)

@exo-cedric so for the sake of argument, I've just tested this scenario: you have to set the `exoscale-ops` account as default account in your `exoscale.toml` CLI configuration file, then what happens is:

```console
$ ./bin/exo ssh
# 15 seconds later...
Display all 594 possibilities? (y or n)
```

exo-cedric:
> ```
> # 15 seconds later...
> ```

This is what I was weary about; thanks for the quick test! :-)

PR #275, opened 2020-08-06
fix: api

yoan-adfinis:
equivalent is the adjective here


falzm:
https://www.macmillandictionary.com/dictionary/british/equivalent_2

yoan-adfinis:
Which implied that _command_is not the adjective but the noun. Have a nice day.

PR #276, opened 2020-08-10
README: config path

raphink:


README.md
```diff
@@ -29,7 +29,7 @@ Running the `exo config` command will guide you through the initial configuratio
 
 You can find your API credentials in the *Account* section of the [Exoscale Console](https://portal.exoscale.com/account/profile/api).
 
-The configuration file and all assets created during `exo` operations will be saved in the `.exoscale` folder inside your home directory.
+The configuration file and all assets created during `exo` operations will be saved in the `.config/exoscale` folder inside your home directory.
```
falzm, line 32: Well, it's a bit more complicated than that... In practice, it depends on the OS: on Linux and UNIX-like systems it's `$HOME/.config/exoscale`, on macOS it's `~/Library/Application Support/exoscale/`, on MS Windows I have no idea where... As if it weren't complicated enough, we _also_ support the legacy location `$HOME/.exoscale`. So if your change doesn't help everybody, I can't accept it as is I'm afraid ¯\\\_(ツ)_/¯

SebastienPittet: On Windows :

%USERPROFILE%\.exoscale\exoscale.toml

raphink: Oh well. I was just surprised because `exo` generated its conf in `~/.config/exoscale` on my machine. Thanks for the explanation.

falzm: @raphink agreed, this situation is not ideal. We should document this more extensively in the README, but in any case this PR won't cut it. Thank you for trying, though :)

falzm:
FYI, README has been updated in a14117d.

raphink:
Thanks!

PR #280, opened 2020-08-18
Write async operation output to stderr

falzm:
This change prints the output of asynchronous operations to `stderr`
instead of `stdout` by default, so that users can discard those using
`2>/null` and still use the `--output|-O` flag to retrieve information
after operation completion.

### DIFF
--- a/cmd/request.go
+++ b/cmd/request.go
@@ -40,6 +40,7 @@ func asyncTasks(tasks []task) []taskResponse {
 	maximum := 1 << 30
 	var taskWG sync.WaitGroup
 	progress := mpb.NewWithContext(gContext,
+		mpb.WithOutput(os.Stderr),
 		mpb.WithWaitGroup(&taskWG),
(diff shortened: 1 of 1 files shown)


falzm:
Before:

```console
$ exo instancepool create test --service-offering Small --template 'Linux Ubuntu 20.04 LTS 64-bit' --output-template '{{ .ID }}' 2>/dev/null
Creating instance pool "test" ⠧                                                                                done
68666ded-2741-9b28-0b0c-8a6704364a5d
```

After:

```console
$ exo instancepool create test --service-offering Small --template 'Linux Ubuntu 20.04 LTS 64-bit' --output-template '{{ .ID }}' 2>/dev/null
4e8aa910-e5d9-19dc-3948-8c69f132746b
```

falzm:
@exo-cedric makes sense, the sensible thing to do IMO would be to add a `--progress` flag disabled by default to have a "clean" output if the user wants to keep stderr content relevant, but since we don't have meaningful progress reporting it's a bit misleading... We might as well remove the spinner completely, in the end.

PR #281, opened 2020-08-26
instancepool: add support for IPv6 

falzm:
This change adds support for IPv6 enabling on Instance Pools.


falzm:
Note that, as for the rest of the tooling, it is not possible to disable IPv6 once it's enabled.

pierre-emmanuelJ (approved):

PR #283, opened 2020-09-02
eip: add support for new TLS options

pyr:


cmd/eip_update.go
```diff
@@ -53,16 +53,26 @@ var eipUpdateCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
+		tlsSni, err := cmd.Flags().GetString("healthcheck-tls-sni")
```
falzm, line 56: Per convention we maintain acronyms in uppercase (when they're not prefixing the name, as it would export them out of the Go package):

```suggestion
		tlsSNI, err := cmd.Flags().GetString("healthcheck-tls-sni")
```

pierre-emmanuelJ (approved):
LGTM thanks, maybe you can run a `go mod tidy` ;)

illi-j (approved):

falzm (changes requested):
Looks good, but can you please also update the relevant CLI commands descriptions to include `https`? (e.g. here https://github.com/exoscale/cli/blob/0945f66b7173fa083f7e7acaeab4e776aeef20a4/cmd/eip_create.go#L108)

falzm (approved):

PR #286, opened 2020-09-17
documentation fixes for exo config

retrack:
This is a proposal to align documentation to best practices regarding API credentials:

* redirect users to IAM instead of ACCOUNT section which is deprecated
* point to the correct configuration path

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,16 +30,28 @@ Upon successful compilation, the resulting `exo` binary is stored in the `bin/` 
 
 Running the `exo config` command will guide you through the initial configuration.
 
-You can find your API credentials in the *Account* section of the [Exoscale Console](https://portal.exoscale.com/account/profile/api).
+You can create and find API credentials in the *IAM* section of the [Exoscale Console](https://portal.exoscale.com/iam/api-keys).
 
 The configuration file and all assets created during `exo` operations will be saved in the following location:
 
 | OS | Location |
 |:--|:--|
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,16 +30,28 @@ Upon successful compilation, the resulting `exo` binary is stored in the `bin/`
...
 
 The configuration file and all assets created during `exo` operations will be saved in the following location:
 
 | OS | Location |
 |:--|:--|
-| GNU/Linux, *BSD | `$HOME/.exoscale` |
-| macOS | `$HOME/Library/Application Support/exoscale` |
-| Windows | `%USERPROFILE%\.exoscale` |
+| GNU/Linux, *BSD | `$HOME/.config/exoscale/` |
+| macOS | `$HOME/.config/exoscale/` or `$HOME/Library/Application Support/exoscale/` |
```
falzm, line 40: I'm not sure to agree on this one: we should only list one official location, not all the legacy still supported.

retrack: @falzm agreed, but when your access does not work for some reason, knowing where to look is useful. Do give me the supported location per OS family with the official and deprecated ones. Then I can split this in 2 tables, ok ?

retrack: Anyhow I just confirmed that $HOME/.exoscale is wrong `exo config` initializes on $HOME/.config/exoscale

falzm (approved):
Thanks @retrack

PR #291, opened 2020-10-22
fix(typo): shortname of `snapshot export` should be `Export snapshot` instead of `export snapshot`

arnaudgeiser:


falzm (approved):
Thanks @arnaudgeiser.

PR #292, opened 2020-10-22
lab: add "coi" command

falzm:
This change introduces a new `exo lab coi` command, to deploy
Container-Optimized Instances.

### DIFF
--- a/cmd/coi.go
+++ b/cmd/coi.go
@@ -1,0 +1,249 @@
+package cmd
+
+import (
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/exoscale/egoscale"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+)
+
+const coiTemplateName = "Container-Optimized Instance"
+
+var coiCmd = &cobra.Command{
+	Use:   "coi <name>",
+	Short: "Deploy a Container-Optimized Instance",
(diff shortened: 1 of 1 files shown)


falzm:
[ch20397]

gabrielexoscale (approved):

illi-j (approved):
being able to specify environments when user doesn't choose the docker-compose option might be nice. What do you think?

otherwise LGTM! nice feature 👍

falzm:
> being able to specify environments when user doesn't choose the docker-compose option might be nice. What do you think?

WDYM by *environment* ? 🤔

illi-j:
> WDYM by _environment_ ?

docker run -e 

and the volumes too:
docker run -v

falzm:
Ah, environment variables. Well IMHO this falls into the "advanced usage" category, which prompts a proper Docker Compose file based workflow in order to preserve the simplicity of the image based use case.

cmd/coi.go
```diff
@@ -0,0 +1,249 @@
...
+		if err != nil {
+			return err
+		}
+
+		serviceOffering, err := getServiceOfferingByNameOrID(so)
+		if err != nil {
+			return err
+		}
+
+		template, err := getTemplateByNameOrID(zone.ID, coiTemplateName, "community")
```
pierre-emmanuelJ, line 114: IMHO, You should use `Fedora CoreOS`, it's natively container optimized 😉

pierre-emmanuelJ: Well, just checked, docker-compose is not installed by default on it 😕 only Docker

PR #297, opened 2020-11-16
vm: support custom output for "create" command

### LINKED ISSUE #294: exo vm create ignores --output-format flag
```
joggie@Joggies-PC:~$ exo vm create test -O json
Creating private SSH key
Deploying "test" ⠋                                                                                done

What to do now?

1. Connect to the machine

        exo ssh test

2. Put the following SSH configuration into ".ssh/config"

Host test
        HostName xxx
        User ubuntu
        IdentityFile /home/joggie/.config/exoscale/instances/xxx/id_rsa

Tip of the day:
        You're the sole owner of the private key. Be cautious with it.
```

The output should be some kind of json to work with, but its a string.
For automation with ansible/molecule I have to parse the string with regex... if something changes in the string the regex may fail.
Please allow the output-format flag for vm create

thanks

falzm:
This change refactors the `vm create` command implementation to support
global `--output` flag:

```command
$ go run main.go vm create blah --disk 10 -O json | jq .
Creating single-use SSH key
Deploying "blah" ⠴                                       done
{
  "id": "d5c46409-5c4e-4269-995a-2395bbca8128",
  "name": "blah",
  "creation_date": "2020-11-16T09:11:58+0100",
  "size": "Medium",
  "disk_size": "10 GiB",
  "template": "Linux Ubuntu 20.04 LTS 64-bit",
  "zone": "ch-gva-2",
  "state": "Running",
  "ip_address": "159.100.241.155",
  "username": "ubuntu",
  "ssh_key": "/Users/marc/Library/Application Support/exoscale/instances/d5c46409-5c4e-4269-995a-2395bbca8128/id_rsa",
  "security_groups": [
    "default"
  ]
}
```

Besides, the command now only supports creating a single instance to make it more consistent with the rest of the commands.

Fixes #294.

### DIFF
--- a/cmd/kube_create.go
+++ b/cmd/kube_create.go
@@ -162,7 +162,8 @@ sudo -E DEBIAN_FRONTEND=noninteractive apt-get install -y kubelet=${PKG_VERSION}
 	kubeadm=${PKG_VERSION} \
 	kubectl=${PKG_VERSION}
 sudo apt-mark hold kubelet kubeadm kubectl`,
-	}, {
+	},
+	{
 		name: "Kubernetes cluster node initialization",
 		command: `\
 set -xe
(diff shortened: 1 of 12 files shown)


falzm:
[ch22108]

cmd/kube_create.go
```diff
@@ -537,7 +534,7 @@ func (c *sshClient) scp(src, dst string) error {
 		}
 	}
 
-	return ioutil.WriteFile(dst, buf.Bytes(), 0600)
+	return ioutil.WriteFile(dst, buf.Bytes(), 0o600)
```
illi-j, line 537: I think the "o" is not expected

illi-j:
we could choose to not specify a "name", as on the Exoscale console and generate one (VM-<vmID>)

illi-j (approved):
except my previous comments, LGTM!

cmd/kube_create.go
falzm: It is: https://www.infoworld.com/article/3408740/golang-113-adds-number-literal-prefixes.html

illi-j: my bad! :) thx

exo-cedric (approved):
afaict, lgtm

PR #304, opened 2020-12-16
sos upload: always send content md5

brutasse:
For multipart uploads, this triggers this code path:

https://github.com/minio/minio-go/blob/v6.0.57/api-put-object-streaming.go#L278

For non-multipart, this code path:

https://github.com/minio/minio-go/blob/v6.0.57/api-put-object-streaming.go#L384

This results in put requests being done with a `Content-MD5` header,
allowing the client and the backend to agree on data integrity.


LaCap (approved):
Thanks @brutasse

falzm (approved):
LGTM 👍

PR #307, opened 2021-01-11
Update "sks" commands

falzm:
* New command `exo sks nodepool evict`
* New command `exo sks versions`
* `exo sks create`: use version 1.20.0 by default


sgmac (approved):
LGTM Nice work!

mcorbin (approved):

PR #314, opened 2021-02-19
sks: nodepool: add "list" command

falzm:
This change introduces a new `exo sks nodepool list` command.

### DIFF
--- a/cmd/sks_list.go
+++ b/cmd/sks_list.go
@@ -54,8 +54,8 @@ func listSKSClusters(zone string) outputter {
 	defer close(res)
 
 	go func() {
-		for nlb := range res {
-			out = append(out, nlb)
+		for cluster := range res {
+			out = append(out, cluster)
 		}
 	}()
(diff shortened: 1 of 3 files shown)


aarongroom (approved):

cmd/sks_list.go
```diff
@@ -65,18 +65,18 @@ func listSKSClusters(zone string) outputter {
...
+				Name: cluster.Name,
 				Zone: zone,
 			}
 		}
 
 		return nil
 	})
 	if err != nil {
-		fmt.Fprintf(os.Stderr,
+		_, _ = fmt.Fprintf(os.Stderr,
```
zyegfryed, line 79: Out of curiosity, why do we have to capture to dead variables the output of `fmt.Fprintf`?

zyegfryed (approved):

cmd/sks_list.go
falzm: Otherwise my overzealous IDE complains that the returned error is not checked 😒

PR #323, opened 2021-03-25
sks: kubeconfig: add support for exec credential mode

### LINKED ISSUE #322: Support client-go credential plugins for authentication to SKS
Currently to connect to SKS we have to generate a full kubeconfig file with `exo sks kubeconfig ... > kubeconfig.yaml`.
It would be great to be able to have a static kubeconfig file that calls the `exo` binary to retrieve the token dynamically.
Reference: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins

falzm:
This change adds support for usage of the `exo sks kubeconfig` command
in credential plugin mode, where it returns a kubeconfig `user` data
dynamically to the `kubectl` CLI.

Fixes #322.

### DIFF
--- a/cmd/sks_kubeconfig.go
+++ b/cmd/sks_kubeconfig.go
@@ -2,11 +2,13 @@ package cmd
 
 import (
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"time"
 
 	exoapi "github.com/exoscale/egoscale/v2/api"
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
 )
 
 var sksKubeconfigCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/sks_kubeconfig.go
```diff
@@ -33,7 +35,47 @@ Example usage:
...
+    - name: my-sks-cluster
+      context:
+        cluster: my-sks-cluster
+        user: exo@my-sks-cluster
+    current-context: my-sks-cluster
+
+Notes:
+
+* The "exo" CLI binary must be installed in a directory listed your PATH shell
+  environment variable.
```
zyegfryed, line 73:
```suggestion
* The "exo" CLI binary must be installed in a directory listed in your PATH
  shell environment variable.
```

zyegfryed (approved):
lgtm 👍🏻

cmd/sks_kubeconfig.go
```diff
@@ -33,7 +35,47 @@ Example usage:
...
+        certificate-authority-data: **BASE64-ENCODED CLUSTER CERTIFICATE**
+        server: https://153fcc53-1197-46ae-a8e0-ccf6d09efcb0.sks-ch-gva-2.exo.io:443
+    users:
+    - name: exo@my-sks-cluster
+      user:
+        # The "exec" section replaces "client-certificate-data"/"client-key-data"
+        exec:
+          apiVersion: "client.authentication.k8s.io/v1beta1"
+          command: "exo"
+          args: ["sks", "kubeconfig", "my-sks-cluster", "--exec-credential", "user"]
```
mcorbin, line 62: I think we should also add the zone here.

mcorbin (approved):

cmd/sks_kubeconfig.go
falzm: Good point, added 👍

PR #326, opened 2021-03-30
storage: bug fixes

falzm:
## `storage`: fix prefix-based listing logic

This change fixes the listing logic used in the `exo storage` commands,
where previously all objects matching the specified prefix were
processed – which is not wanted when the prefix points to a discrete
file. The updated logic returns only the object matching the specified
prefix if the prefix doesn't end with a trailing separator ("/").

## `storage download`: fix destination file path

This change fixes the behavior of the `exo storage download` command,
where previously downloading a file located in a prefix would re-create
the prefix directory tree locally. The updated logic now stores the
specified file under its base name in the specified local directory.

## `storage download`: fix relative destination paths

This change fixes a bug causing a "file already exists" error message
returned when users pass a destination as relative path (e.g. ".", "..",
"../..") without trailing slash.

### DIFF
--- a/cmd/storage.go
+++ b/cmd/storage.go
@@ -175,6 +175,14 @@ func (c *storageClient) forEachObject(bucket, prefix string, recursive bool, fn 
 				continue
 			}
 
+			// If the prefix doesn't end with a trailing prefix separator ("/"),
+			// consider it as a single object key and match only one exact result
+			// (except in recursive mode, where the prefix is expected to be a
+			// "directory").
+			if !recursive && !strings.HasSuffix(prefix, "/") && aws.ToString(o.Key) != prefix {
+				continue
+			}
+
 			o := o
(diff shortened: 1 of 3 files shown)


sgmac (approved):

cmd/storage.go
```diff
@@ -175,6 +175,14 @@ func (c *storageClient) forEachObject(bucket, prefix string, recursive bool, fn
 				continue
 			}
 
+			// If the prefix doest't end with a trailing prefix separator ("/"),
```
bcachet, line 178:
```suggestion
			// If the prefix doesn't end with a trailing prefix separator ("/"),
```

bcachet (approved):

cmd/storage.go
falzm: Good catch, fixed – thanks!

PR #332, opened 2021-04-16
env: support variables unsetting

falzm:
This change adds a new flag `-u|--unset` to the `exo env` command,
allowing removal of the environment variables set by the `exo env`
command.

### DIFF
--- a/cmd/environment.go
+++ b/cmd/environment.go
@@ -6,6 +6,29 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var envCmd = &cobra.Command{
+	Use:    "env",
+	Hidden: true,
+	Run: func(cmd *cobra.Command, _ []string) {
+		vars := map[string]string{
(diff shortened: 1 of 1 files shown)


cmd/environment.go
```diff
@@ -6,6 +6,29 @@ import (
...
+			"EXOSCALE_API_ENVIRONMENT": gCurrentAccount.Environment,
+		}
+
+		unset, _ := cmd.Flags().GetBool("unset")
+
+		for k, v := range vars {
+			if unset {
+				fmt.Printf("unset %s\n", k)
+			} else {
+				fmt.Printf("export %s=%q\n", k, v)
```
exo-cedric, line 26:
```suggestion
				fmt.Printf("export %s='%q'\n", k, v)
```

falzm: @exo-cedric `%q` already encloses the value with quotes, it's equivalent to `"%q"`.

exo-cedric (approved):

PR #337, opened 2021-05-26
Add autocompletion generation for more shells

falzm:
This change adds shell autocompletion support for Fish, PowerShell &
Zsh.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -72,6 +72,8 @@ brews:
       bin.install "exo"
       man1.install Dir["manpage/exo*.1"]
       bash_completion.install "contrib/completion/bash/exo"
+      fish_completion.install "contrib/completion/fish/exo"
+      zsh_completion.install "contrib/completion/zsh/exo"
 
 dockers:
   - goos: linux
--- a/Makefile
+++ b/Makefile
@@ -45,9 +45,11 @@ manpages: manpage
 
 .PHONY: completions
 completions:
-	mkdir -p contrib/completion/bash
-	$(GO) run -mod vendor completion/main.go
-	mv bash_completion contrib/completion/bash/exo
+	mkdir -p contrib/completion/{bash,fish,powershell,zsh}
+	$(GO) run -mod vendor completion/main.go bash ; mv bash_completion contrib/completion/bash/exo
+	$(GO) run -mod vendor completion/main.go fish ; mv fish_completion contrib/completion/fish/exo
+	$(GO) run -mod vendor completion/main.go powershell ; mv powershell_completion contrib/completion/powershell/exo
+	$(GO) run -mod vendor completion/main.go zsh ; mv zsh_completion contrib/completion/zsh/exo
 
 .PHONY: clean
 clean::
--- a/completion/main.go
+++ b/completion/main.go
@@ -2,12 +2,35 @@ package main
 
 import (
 	"log"
+	"os"
 
 	"github.com/exoscale/cli/cmd"
 )
 
 func main() {
-	if err := cmd.RootCmd.GenBashCompletionFile("bash_completion"); err != nil {
+	if len(os.Args) != 2 {
+		log.Fatalf("usage: %s (bash|fish|powershell|zsh)", os.Args[0])
+	}
+
+	var err error
+	switch os.Args[1] {
+	case "bash":
+		err = cmd.RootCmd.GenBashCompletionFile("bash_completion")
+
+	case "fish":
+		err = cmd.RootCmd.GenFishCompletionFile("fish_completion", true)
+
+	case "powershell":
+		err = cmd.RootCmd.GenPowerShellCompletionFile("powershell_completion")
+
+	case "zsh":
+		err = cmd.RootCmd.GenZshCompletionFile("zsh_completion")
+
+	default:
+		log.Fatalf("unsupported shell %q", os.Args[1])
+	}
+
+	if err != nil {
 		log.Fatal(err)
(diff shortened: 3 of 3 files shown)


zyegfryed (approved):
Can't wait to test it with my `zsh` shell!

.goreleaser.yml
```diff
@@ -72,6 +72,8 @@ brews:
       bin.install "exo"
       man1.install Dir["manpage/exo*.1"]
       bash_completion.install "contrib/completion/bash/exo"
+      fish_completion.install "contrib/completion/fish/exo"
+      zsh_completion.install "contrib/completion/zsh/exo"
```
retrack, line 76:
```suggestion
      zsh_completion.install "contrib/completion/zsh/_exo"
```

for me adding the file as exo does not work, but installing it manually as _exo works. I do not know if the goreleaser does add the underscore implicitly while installing

Makefile
```diff
@@ -45,9 +45,11 @@ manpages: manpage
...
 .PHONY: completions
 completions:
-	mkdir -p contrib/completion/bash
-	$(GO) run -mod vendor completion/main.go
-	mv bash_completion contrib/completion/bash/exo
+	mkdir -p contrib/completion/{bash,fish,powershell,zsh}
+	$(GO) run -mod vendor completion/main.go bash ; mv bash_completion contrib/completion/bash/exo
+	$(GO) run -mod vendor completion/main.go fish ; mv fish_completion contrib/completion/fish/exo
+	$(GO) run -mod vendor completion/main.go powershell ; mv powershell_completion contrib/completion/powershell/exo
+	$(GO) run -mod vendor completion/main.go zsh ; mv zsh_completion contrib/completion/zsh/exo
```
retrack, line 52:
```suggestion
	$(GO) run -mod vendor completion/main.go zsh ; mv zsh_completion contrib/completion/zsh/_exo
```

retrack (commented):
I cannot get it to work properly and get the below on zsh: (-> is tab)

`acr@mbp13ar-3 zsh % exo li->
_arguments:comparguments:325: invalid option definition: (-C --config)-C[Specify an alternate config file [env EXOSCALE_CONFIG]]:`

did you manage to get is work ?

falzm:
@retrack indeed Zsh seems to require completion files to begin with an underscore.

falzm:
@zyegfryed can you confirm it works with your zsh shell?

retrack:
> @zyegfryed can you confirm it works with your zsh shell?

@falzm can you remind me how to run a local release with goreleaser? maybe using the output of that will help. I tested doing `make build` and `make completions` and then moving the file around. But there might be more to it and I am not a zsh user, good old bash and a mile long .bash_profile ...

zyegfryed:
> > @zyegfryed can you confirm it works with your zsh shell?
> 
> @falzm can you remind me how to run a local release with goreleaser? maybe using the output of that will help. I tested doing `make build` and `make completions` and then moving the file around. But there might be more to it and I am not a zsh user, good old bash and a mile long .bash_profile ...

I'm doing the same and having the exact some behavior as you. I also remove the `.zcompdump` files on my home directory (`rm .zcompdump*`), but still the same issue with wrong argument...
Will continue to try to make it work and will keep you update don my progress.

falzm:
Same, not a `zsh` user so I'm not sure what I'm doing wrong here... 

@retrack you can run `goreleaser build --snapshot --rm-dist` at the root of the source tree.

zyegfryed:
@falzm  @retrack OK, I understand what's the issue: in our commands description we sometime refer to a way to define an option via an environment variable, for example, the [`config`](https://github.com/exoscale/cli/blob/master/cmd/root.go#L111) option. To do so, we surround the environment variable with brackets in the description – i.e. `[env EXOSCALE_CONFIG]`, which breaks the [`_autocomplete`](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Completion-Functions) function.
A quick fix would be to use parentheses instead. Doable?

zyegfryed:
^^ an alternate solution will be to escape the bracket from the description. But I guess we'll need to modify the underlying library ([`cobra`](https://github.com/spf13/cobra)) to do so, wouldn't we?

falzm:
> ^^ bumping cobra to [v1.1.0](https://github.com/spf13/cobra/releases/tag/v1.1.0) should fix the issue since a revamped auto-completion for zsh was rolled out in that release.

@zyegfryed cool, good thing that I have an on-going branch where I've upgraded Cobra to the latest version then. Thank you for looking into this!

falzm:
@zyegfryed I've updated the branch, can I ask you for a new test?

zyegfryed:
> @zyegfryed I've updated the branch, can I ask you for a new test?

Sure. Test coming...

zyegfryed:
So, completion works as expected for the first level commands/options, but fails when going deeper:

```shell
~ ❯ exo v<TAB>                                                                                                                                                                                                                                                                                                                                                    
version  -- Print the version of exo
vm       -- Compute instances management

~ ❯ exo --c<TAB>
~ ❯ exo --config
```

but, this will fail:

```shell
~ ❯ exo vm cre
_arguments:comparguments:325: invalid option definition: (-C --config)-C[Specify an alternate config file [env EXOSCALE_CONFIG]]:
```

zyegfryed:
^^ forget my previous comment, I forgot to delete the completion history or made another mistake will trying it, and after cleaning and re-installing it works as expected! GG! 🎉

zyegfryed (approved):

PR #346, opened 2021-06-18
lab: db: fix crash during types list/show

falzm:
This change fixes as crashing bug occuring during `exo lab db types
list|show` commands.


retrack (approved):
tested working on Darwin.

PR #368, opened 2021-08-13
db: new "ca-certificate" command + connection info output label

falzm:
##  db: add new "ca-certificate" command

This change adds a new `exo lab db get-certificate` command.

##  db: show: expose service-specific connection info

This change introduces a new `ConnectionInfo` output label to the `exo
lab db show` command, exposing service-specific connection info.

##  db: show: expose service components

This change introduces a new `Components` output label to the `exo
lab db show` command, exposing service-specific components.

### DIFF
--- a/cmd/db_cacertificate.go
+++ b/cmd/db_cacertificate.go
@@ -1,0 +1,45 @@
+package cmd
+
+import (
+	"fmt"
+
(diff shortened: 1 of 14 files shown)


lord2y (approved):
not a golang expert but I tested first hand this code and it's work so lgtm 😋

PR #391, opened 2021-10-01
compute: instance: raise default timeout

falzm:
This change raises the default timeout for Compute instance operations
to 10 minutes.


falzm:
[ch35064]

PR #393, opened 2021-10-01
Fix client custom HTTP User-Agent header

falzm:
This change fixes a bug causing the client custom HTTP User-Agent header
not to be set on outgoing API requests.


falzm:
[ch35087]

PR #395, opened 2021-10-08
Promote DBaaS-related commands to GA

falzm:
This change refactors the `exo lab database` commands following recent
Exoscale public API changes. Commands are now promoted to GA in the `exo
database` section.


falzm:
[ch35089]

PR #397, opened 2021-10-15
fix(sos): Calculate appropriate partsize when uploading

wdullaer:
Due to the library which renders the progress bar, the AWS sdk cannot
compute the correct partsize. This means that the maximum file size
which can be uploaded by the CLI is 50 GB: 5 MB x 10 000 parts, because
the AWS SDK has an fixed max of 10k parts.

The change here explicitly calculates and sets a partsize, based on the
detected size of the object which is being uploaded.

An exo binary with this change applied managed to upload an object of
250 GB.

[sc-36086]

### DIFF
--- a/cmd/storage_upload.go
+++ b/cmd/storage_upload.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


pyr (approved):

cmd/storage_upload.go
```diff
@@ -322,3 +335,35 @@ func (c *storageClient) uploadFile(bucket, file, key, acl string) error {
 
 	return err
 }
+
+func estimatePartSize(f *os.File) (int64, error) {
```
falzm, line 339: Since we're having all the CLI commands sharing the same package namespace, I usually avoid generic names for functions. Since it's S3-specific, I'd suggest either to rename it to `estimateStoragePartSize()` or to implement it as a method on the `*storageClient` receiver.

falzm (approved):
Minor function naming update, this looks good to me. Thanks a lot @wdullaer 👍

PR #442, opened 2022-03-11
Add nodepool flag for linbit

dramaturg:
This adds handling addon flags indentically to what we do for clusters.

### DIFF
--- a/cmd/sks_nodepool_add.go
+++ b/cmd/sks_nodepool_add.go
@@ -9,10 +9,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var (
-	sksNodepoolAddonLinbit = "linbit"
-)
-
 type sksNodepoolAddCmd struct {
 	cliCommandSettings `cli-cmd:"-"`
 
--- a/cmd/sks_nodepool_add.go
+++ b/cmd/sks_nodepool_add.go
@@ -68,23 +68,16 @@ func (c *sksNodepoolAddCmd) cmdRun(_ *cobra.Command, _ []string) error {
 		return fmt.Errorf("error retrieving cluster: %w", err)
 	}
 
-	addOns := map[string]struct{}{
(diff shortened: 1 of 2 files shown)


exo-cedric:
May I suggest we'd rather introduce a --partitioning (or --disk-part or ...) parameter, allowing to cover as many partitioning use cases as we might want (default, linstore, luks, etc.)

cmd/sks_nodepool_add.go
```diff
@@ -63,6 +68,24 @@ func (c *sksNodepoolAddCmd) cmdRun(_ *cobra.Command, _ []string) error {
...
+
+		if len(addOns) > 0 {
+			list := make([]string, 0)
+			for k := range addOns {
+				list = append(list, k)
+			}
+			v = &list
+		}
+		return
+	}()
```
7fELF, line 87:
```suggestion
	addOns := map[string]bool{
		"linbit": c.Linbit,
	}

	nodepool.AddOns = &[]string{}
	for k, v := range addOns {
		if v {
			*nodepool.AddOns = append(*nodepool.AddOns, k)
		}
	}
```

(tested)

cmd/sks_nodepool_add.go
```diff
@@ -9,6 +9,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	sksNodepoolAddonLinbit = "linbit"
+)
+
```
7fELF, line 15:
```suggestion
```

7fELF (approved):

dramaturg:
> May I suggest we'd rather introduce a --partitioning (or --disk-part or ...) parameter, allowing to cover as many partitioning use cases as we might want (default, linstore, luks, etc.)

We can still do that if we have multiple options at some point in time.

PhilippeChepy (approved):

PR #444, opened 2022-03-31
Fix template import ignored parameters

PhilippeChepy:
Some user-supplied arguments are ignored when building a template from a snapshot.
This should fix [sc-44525]

I took the opportunity to build it using 1.17


PhilippeChepy:
Now handling correctly "--disable-password" field

pyr (approved):
:rocket:

exo-cedric (approved):

PR #446, opened 2022-04-07
makefile: allow supplying an alternative OpenAPI document

pyr:
The recent modification to the Makefile enforces downloading a
source document which is inconvenient while working on new API
functionality.

This patch allows downloading by default but will also let
developers supply their own document if necessary.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,4 +20,4 @@ jobs:
     - name: Run tests and attempt building
       run: |
         git submodule update --init --recursive go.mk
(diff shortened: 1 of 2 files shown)


exo-cedric (approved):
(notwithstanding CI tests pass)

pyr:
The linter uses complains by itself that it is deprecated, I took the opportunity to switch to the recommended one in the warning message (**revive**) which fixes the linting failures

7fELF (approved):

PR #449, opened 2022-04-18
Predrag/sc 45726/cli dbaas add update migration support

kobajagi:
Adds support for running migration when creating/updating supported dbaas service. Services supporting migrations are mysql, pg and redis.
Also fixes a panic when using `--maintenance-dow` and `--maintenance-time` flags in `dbaas create` command.


kobajagi:
Added support for `--mysql-binlog-retention-period` as well.

jessicatoscani (approved):

PhilippeChepy (approved):

PR #450, opened 2022-04-18
instance_list, instance_snapshot_list: fix race condition

kobajagi:
Fixes race condition in instance/snapshot list command. This is what was causing the missing entries in the list command: results were sometimes printed before goroutine pushed all data to output channel.

### DIFF
--- a/cmd/instance_list.go
+++ b/cmd/instance_list.go
@@ -59,7 +59,7 @@ func (c *instanceListCmd) cmdRun(_ *cobra.Command, _ []string) error {
 
 	out := make(instanceListOutput, 0)
 	res := make(chan instanceListItemOutput)
-	defer close(res)
+	done := make(chan struct{})
 
 	instanceTypes := make(map[string]*egoscale.InstanceType) // For caching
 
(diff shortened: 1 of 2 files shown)


cmd/dbaas_list.go
```diff
@@ -56,12 +56,13 @@ func (c *dbaasServiceListCmd) cmdRun(_ *cobra.Command, _ []string) error {
 
 	out := make(dbaasServiceListOutput, 0)
 	res := make(chan dbaasServiceListItemOutput)
-	defer close(res)
+	done := make(chan struct{})
```
bcachet, line 59: The `done` chan is never closed. 
Is there any good reason not do it ?

bcachet (approved):
Nice catch
Using `chan` to notify that the _print to stdout_ operation is finished is a solid
I am just wondering why we do not close the `done` channel we are using for synchro (it is not really an issue, as the program will finish once the command is finished)

cmd/dbaas_list.go
PhilippeChepy: It is OK to leave a channel open. They are garbage collected if no longer in use.
Closing res is required as it's a condition to exit the loop L62.

7fELF (approved):

cmd/dbaas_list.go
kobajagi: List command process will exit fast (after printing the results) so channel will not live long anyway and closing channel does not have any impact on the workflow. It would be correct to close the channel, but it this particular case I'd rather have one less line in the code.

PR #451, opened 2022-04-29
Compute: Do not compress user-data by default

pierre-emmanuelJ:
For some consistency with the Web portal, we should not compress the user-data by default.

In the context of the Fedora CoreOS template, Ignition accept only not compressed user-data:
https://github.com/coreos/ignition/pull/1345


illi-j (approved):

pierre-emmanuelJ:
> https://github.com/exoscale/cli/pull/451/commits/2ca23907ddaf001eb208a9b0d5d75d50a3b9fc9c

This is enough to fallback on a non Gzipped data decoding:
https://github.com/exoscale/cli/blob/master/cmd/vm.go#L144-L147

aureliar8 (approved):

PR #459, opened 2022-06-14
Opensearch integration in CLI

7fELF:


cmd/dbaas.go
```diff
@@ -153,3 +154,18 @@ func dbaasShowSettings(settings map[string]interface{}) {
 		t.Append(row)
 	}
 }
+
+func dbaasGetType(ctx context.Context, name, zone string) (string, error) {
+	dbs, err := cs.ListDatabaseServices(ctx, zone)
+	if err != nil {
+		return "", fmt.Errorf("failed to retrieve database type: %s", err)
```
kobajagi, line 161: It doesn't cost us anything to wrap errors (`%w`). Might come in handy in the future.

kobajagi, line 161: It doesn't cost us anything to wrap errors (`%w`). Might come in handy in the future.

cmd/dbaas_show_opensearch.go
```diff
@@ -0,0 +1,302 @@
...
+				Message: wordwrap.WrapString(n.Message, 50),
+				Type:    string(n.Type),
+			})
+		}
+	}
+	return &out, nil
+}
+
+func opensearchShowBackups(db *oapi.DbaasServiceOpensearch) (outputter, error) {
+	out := make(dbServiceBackupListOutput, 0)
```
kobajagi, line 185: It's possible to allocate exact capacity with little reordering:
```
if db.Backups == nil {
  return nil, nil //or return &dbServiceBackupListOutput{}, nil
}
out := make(dbServiceBackupListOutput, 0, len(db.Backups))
for _, b := range *db.Backups {
...
```

7fELF: Sure, it probably won't have any significant performance impact, but it looks better! (Took this part from [here](https://github.com/exoscale/cli/blob/596501c3ed402bebb406b948646d6e37c2fe9cde/cmd/dbaas_show_kafka.go#L153))

cmd/dbaas.go
7fELF:
```suggestion
		return "", fmt.Errorf("failed to retrieve database type: %w", err)
```

cmd/dbaas_update_opensearch.go
```diff
@@ -0,0 +1,152 @@
...
+	if cmd.Flags().Changed(mustCLICommandFlagName(c, &c.OpensearchIndexPatterns)) {
+		db.IndexPatterns = &[]struct {
+			MaxIndexCount    *int64                                                                  `json:"max-index-count,omitempty"`
+			Pattern          *string                                                                 `json:"pattern,omitempty"`
+			SortingAlgorithm *oapi.UpdateDbaasServiceOpensearchJSONBodyIndexPatternsSortingAlgorithm `json:"sorting-algorithm,omitempty"`
+		}{}
+
+		err := json.Unmarshal([]byte(c.OpensearchIndexPatterns), db.IndexPatterns)
+		if err != nil {
+			return fmt.Errorf("failed to decode Opensearch index patterns JSON: %s", err)
```
7fELF, line 125:
```suggestion
			return fmt.Errorf("failed to decode Opensearch index patterns JSON: %w", err)
```

arapaho (approved):

PR #460, opened 2022-06-20
Migrate DNS commands to Public API

kobajagi:


kobajagi:
Additional improvement is that DNS domain argument to a command can be either name or ID (same with domain record argument).

cmd/dns_list.go
```diff
@@ -57,37 +51,33 @@ Supported output template annotations: %s`,
 }
 
 func listDomains(filters []string) (outputter, error) {
-	domains, err := csDNS.GetDomains(gContext)
+	domains, err := cs.ListDNSDomains(gContext, gCurrentAccount.DefaultZone)
```
7fELF, line 54:
```suggestion
	ctx := exoapi.WithEndpoint(gContext, exoapi.NewReqEndpoint(gCurrentAccount.Environment, gCurrentAccount.DefaultZone))
	domains, err := cs.ListDNSDomains(ctx, gCurrentAccount.DefaultZone)

```

Without this, requests will fail in other envs.

cmd/dns_list.go
```diff
@@ -57,37 +51,33 @@ Supported output template annotations: %s`,
...
+		o := dnsListItemOutput{
+			ID:   StrPtrFormatOutput(d.ID),
+			Name: StrPtrFormatOutput(d.UnicodeName),
 		}
 
-		if !keep {
-			continue
+		if len(filters) == 0 {
+			out = append(out, o)
+			break
```
7fELF, line 69:
```suggestion
			continue
```

cmd/dns_list.go
```diff
@@ -57,37 +51,33 @@ Supported output template annotations: %s`,
...
-			Name:        d.Name,
-			UnicodeName: d.UnicodeName,
-		})
+		s := strings.ToLower(fmt.Sprintf("%s#%s", o.ID, o.Name))
+
+		for _, filter := range filters {
+			substr := strings.ToLower(filter)
+			if strings.Contains(s, substr) {
+				out = append(out, o)
+				break
```
7fELF, line 78:
```suggestion
				continue
```

cmd/dns_list.go
kobajagi: Ah yes, good catch. I'll fix it.

cmd/dns_list.go
kobajagi: I think it should be `break`, or at least that is how i am interpreting filters feature. This runs in nested loop, once first filter (substring) matches we will add it to output and move to the next domain. If `continue` it will try to match next filter on already added domain and may add it twice.

cmd/dns_list.go
kobajagi: Yup, it should be `continue`, my bad.

cmd/dns_list.go
7fELF: ah yes, right!

7fELF (approved):

PR #475, opened 2022-10-15
feat(completion): fig completion added

rama-kairi:


rama-kairi:
image

rama-kairi:
I just published it for my personal use now. 

you can follow the doc - https://fig.io/docs and may extend it further.

pyr:
Hi @rama-kairi,

Thanks for your suggestion! We don't have fig users internally so can't comment on the code but everything looks good. A couple of comments:

- The command is top-level which would be a bit confusing, it should be nested under a separate namespace
- The resulting typescript completion help shouldn't be bundled in the codebase. I'll let the other maintainers weigh-in as well.

kobajagi:
I moved fig command to a dedicated namespace (`integrations`). It's not shown in the help as fig-cobra integration sets hidden attribute on the command (still useful to have a namespace for a 3rd party tools integration).

sauterp (approved):
lgtm

PhilippeChepy (approved):

PR #479, opened 2022-11-29
storage: leverage delimiter when listing objects

kobajagi:
This PR implements delimiters for listing bucket objects (`exo storage list`).

By using a delimiter, listing large buckets will work much faster. Note that delimiter is not used with `--recursive` flag.


zyegfryed (approved):

PhilippeChepy (approved):
:+1:

arnaudgeiser (approved):
Looking good, thanks!

PR #489, opened 2023-02-16
security group output: show instances in security group

sauterp:
We add an "Instances" section to the output of the `securityGroupShowCmd` displaying the name, id, public IP if available and zone of the instances attached to a security group.


utils/utils.go
```diff
@@ -14,3 +18,29 @@ func RandStringBytes(n int) (string, error) {
...
+			if sgID == securityGroupID {
+				instancesInSG = append(instancesInSG, instance)
+			}
+		}
+	}
+
+	return instancesInSG, nil
+}
+
+func GetDefaultStringOrZero(v *string, defaultStr string) string {
```
kobajagi, line 40: Same helper exists here:
https://github.com/exoscale/cli/blob/7ced7f01c89496050a7dc00c80a171c23a775fad/cmd/util.go#L34
I agree package `utils` is a better place for it though.

cmd/security_group_show.go
```diff
@@ -76,6 +85,31 @@ func (o *securityGroupShowOutput) toTable() {
 
 			return buf.String()
 		}
+
+		return "-"
+	}
+
+	formatInstances := func(instances []securityGroupInstanceOutput) string {
+		if len(instances) > 0 {
```
kobajagi, line 93: Just a suggestion, you can flatten the code block if you first add:

```
if len(instances) == 0 {
    return "-"
}
...
```

IMO code is a bit easier to read that way.

jessicatoscani (approved):

kobajagi (approved):

PR #490, opened 2023-02-24
elastic_ip: show names of instances attached to EIP

sauterp:
We show the names of the instances attached to an elastic IP when the user calls `exo c elastic-ip show $SOME_EIP`.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -46,30 +46,12 @@ func GetInstancesInSecurityGroup(ctx context.Context, client *egoscale.Client, s
 }
 
 func GetInstancesAttachedToEIP(ctx context.Context, client *egoscale.Client, elasticIPID, zone string) ([]*v2.Instance, error) {
-	allInstances, err := client.ListInstances(ctx, zone)
+	instances, err := client.ListInstances(ctx, zone, v2.ListInstancesByIpAddress(elasticIPID))
 	if err != nil {
 		return nil, err
 	}
 
-	var instancesAttachedToEIP []*v2.Instance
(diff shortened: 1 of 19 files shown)


utils/utils.go
```diff
@@ -41,6 +45,33 @@ func GetInstancesInSecurityGroup(ctx context.Context, client *egoscale.Client, s
...
 
+func GetInstancesAttachedToEIP(ctx context.Context, client *egoscale.Client, elasticIPID, zone string) ([]*v2.Instance, error) {
+	allInstances, err := client.ListInstances(ctx, zone)
+	if err != nil {
+		return nil, err
+	}
+
+	var instancesAttachedToEIP []*v2.Instance
+	for _, instance := range allInstances {
+		completeInstanceInfo, err := client.GetInstance(ctx, zone, *instance.ID)
```
sauterp, line 56: I'm not sure what to do here. ListInstance returns a slice of Instances which has the field ElasticPIIDs but it's nil even when the instances have an elastic IP. So to get them, I do a GetInstance call here on each instance.
I thought this was a bug, but when I researched it I realized that ListInstances was removed in newer versions of egoscale.

Another option I found was https://github.com/exoscale/cli/blob/master/cmd/eip_show.go#L116
But this command is deprecated.

What's the best way to list instances attached to an EIP?

kobajagi (approved):

PhilippeChepy (approved):

utils/utils.go
PhilippeChepy: TBH, I'm not sure there is "a best way" :)

PR #493, opened 2023-03-31
Remove hardcoded instance snapshot timeout

kobajagi:
Instance snapshot command had a hardcoded timeout of 60 minutes without a way to extend it. This PR removes the hardcoded value. Default value is now 10 minutes (fairly small for snapshot), however timeout override is applied and can easily extend value (env `EXOSCALE_API_TIMEOUT` or config `clientTimeout`).

**edit:** Default Timeout bumped to 20 minutes. Also added descriptive error when timeout is hit on create snapshot command:

```
$ EXOSCALE_API_TIMEOUT=1 go run . c i snapshot create myinstance
 ✔ Creating snapshot of instance "myinstance"... 1m0s
error: Request timeout reached. Snapshot creation is not canceled and might still be running, check the status with : exo c i snapshot list
exit status 1
```

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -88,7 +88,7 @@ const (
 	defaultRunstatusEndpoint  = "https://api.runstatus.com"
 	defaultZone               = "ch-dk-2"
 	defaultOutputFormat       = "table"
-	defaultClientTimeout      = 10
+	defaultClientTimeout      = 20
 )
 
 var configCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


pyr (approved):

aureliar8 (commented):
Maybe we want to simply disable timeouts for snapshot operations?

kobajagi:
> Maybe we want to simply disable timeouts for snapshot operations?

Well having API timeout as a config option one would expect that it applies everywhere. Maybe we can bump the default timeout.

sauterp (approved):

sauterp (approved):

jessicatoscani (approved):

PR #494, opened 2023-03-31
compute: add support for public security groups and password reveal

pyr:
New behavior

- `exo compute security-group list --visibility (public|private)`
- `exo compute instance reveal-password ${ID|NAME}`

With guidance from @kobajagi, thanks!

### DIFF
--- a/cmd/instance_reveal_password.go
+++ b/cmd/instance_reveal_password.go
@@ -1,0 +1,68 @@
+package cmd
+
+import (
+	"errors"
+	"fmt"
+
+	exoapi "github.com/exoscale/egoscale/v2/api"
+	"github.com/spf13/cobra"
+)
+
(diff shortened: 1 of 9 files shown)


aureliar8 (approved):

cmd/instance_reveal_password.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		if errors.Is(err, exoapi.ErrNotFound) {
+			return fmt.Errorf("resource not found in zone %q", c.Zone)
+		}
+		return err
+	}
+
+	pwd, err := cs.RevealInstancePassword(ctx, c.Zone, instance)
+	if err != nil {
+		return nil
```
kobajagi, line 54: We should return error here.

PR #498, opened 2023-04-25
sos: add support for setting the object ownership + refactoring

sauterp:
We add support for setting the object ownership.

As a follow up to this feature, we will add support for versioning. We will need to touch most object storage related commands and in anticipation of this we refactored most of the object storage related code. In particular we separated it into a different package. This should allow us to write better testing later.


PhilippeChepy (approved):
Thank. I didn't check thoroughly every single line of code though.

pkg/globalstate/globalstate.go
```diff
@@ -0,0 +1,11 @@
+package globalstate
+
+import (
+	"github.com/exoscale/egoscale"
+)
+
+var (
+	OutputFormat         string
+	GlobalEgoscaleClient *egoscale.Client
```
kobajagi, line 9: I'm thinking `Global` keyword is unnecessary as the package name (`globalstate`) holds the same information. Just a thought, you can leave it as is.

kobajagi (approved):

PR #502, opened 2023-05-02
CI & build plan improvements

kobajagi:
Minor improvements:

### Makefile

- removes explicit `-mod vendor` (it is default anyway).

### goreleaser

- rename MacOS `unified_binary` to be inline with other packages.

### github action

- Run against 2 latest (supported) Go versions;
- Check the code with `go fmt` and `go vet`;
- Validate dependencies;
- Add `govulncheck` job;
- minor QOL improvements.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,14 +10,33 @@ on:
     - 'v*' # Don't run CI tests on release tags
 
 jobs:
-  build:
+  run:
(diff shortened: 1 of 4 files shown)


sauterp:
some ideas on additional checks we could do:
- linter (perhaps a bigger separate PR is necessary to make all code pass linting)
- formatting (gofumpt)
- test coverage

PR #503, opened 2023-05-05
Add dbaas: grafana

kobajagi:
This PR adds a new database type: grafana.

```
exo dbaas type show grafana
┼────────────────────┼─────────────────────────────┼
│ Name               │ grafana                     │
│ Description        │ Grafana - Metrics Dashboard │
│ Available Versions │                             │
│ Default Version    │ -                           │
┼────────────────────┼─────────────────────────────┼
```

### DIFF
--- a/cmd/dbaas_create_grafana.go
+++ b/cmd/dbaas_create_grafana.go
@@ -1,0 +1,74 @@
+package cmd
+
+import (
+	"fmt"
+	"net/http"
+
+	exoapi "github.com/exoscale/egoscale/v2/api"
+	"github.com/exoscale/egoscale/v2/oapi"
+	"github.com/spf13/cobra"
+)
+
+func (c *dbaasServiceCreateCmd) createGrafana(_ *cobra.Command, _ []string) error {
+	var err error
+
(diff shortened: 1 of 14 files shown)


PhilippeChepy (approved):

cmd/dbaas_create_grafana.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/exoscale/cli/pkg/globalstate"
+	exoapi "github.com/exoscale/egoscale/v2/api"
+	"github.com/exoscale/egoscale/v2/oapi"
+	"github.com/spf13/cobra"
+)
+
+func (c *dbaasServiceCreateCmd) createGrafana(_ *cobra.Command, _ []string) error {
+	var err error
+
+	ctx := exoapi.WithEndpoint(gContext, exoapi.NewReqEndpoint(account.CurrentAccount.Environment, c.Zone))
```
sauterp, line 17: AFAIK it's bad practice to use a global context. IMHO we should stop using it in new code, where possible.

sauterp (approved):

cmd/dbaas_create_grafana.go
kobajagi: I agree it is a bad practice but at the moment global context has CTRL+C listener and i can't replace it locally. I'll tackle the issue in the separate effort.

Issue #492, opened 2023-03-22
Old version 1.22.2 in snap


Ken1Nil:
Current version in snap is 1.22.2 from 2020-12-16.

Latest version v1.66.0
Whould be nice if a newer version could be bumped to snap.

### Reproduce:

```
user@ubuntu:~$ snap info exoscale-cli
name:      exoscale-cli
summary:   The Exoscale command line tool
publisher: Exoscale
store-url: https://snapcraft.io/exoscale-cli
license:   Apache-2.0
description: |
  The Exoscale command line tool.

  This snap allows you to use the "exo" command once installed, your
  gateway to the Exoscale platform.

  Getting started:

  `sudo snap install exoscale-cli`

  `exo config`

  `exo vm create my-first-machine`


  Sources:

  Source code for this software is available on github: https://github.com/exoscale/cli
snap-id: XBC1qevH7VWDkhxj3QN0EyR9uCbG5tDE
channels:
  latest/stable:    v1.22.2 2020-12-16 (38) 10MB -
  latest/candidate: ↑
  latest/beta:      ↑
  latest/edge:      v1.22.2 2020-12-16 (38) 10MB -
```

## Question: Why use snap?
Using Github Actions, need exoscale cli to get values. E.g `exo dbaas list -O json | jq .`

### Workaround 1? Errors
Manual installation according to README.md 
```
$ git clone https://github.com/exoscale/cli
$ cd cli
$ git submodule update --init --recursive go.mk
$ make build
```
This errors because github workers that builds from source doesn't put output into `/bin/exo` or `/usr/bin/exo` but in a sub directory `/home/runner/work/_repository_/_repository_/cli/bin/exo`

### Workaround 2? Not future proof
Could simply download .deb file and install it according to https://community.exoscale.com/documentation/tools/exoscale-command-line-interface/#gnulinux

But this limits versions in the future to a specific static cli version.
```
wget https://github.com/exoscale/cli/releases/download/v1.66.0/exoscale-cli_1.66.0_linux_amd64.deb
sudo dpkg -i exoscale-cli_1.66.0_linux_amd64.deb
``` 

### Solution
Bump versions to snap

or

Change `make build` to actually export the output to `/bin/exo ` when running it on Github Actions workers?

sauterp:
Thanks for reaching out.

In Workaround 1 you can run `go build -o /home/runner/work/_repository_/_repository_/cli/bin/exo` instead of `make build`.

[...]

PR #505, opened 2023-05-10
installation: add a script to install the latest version on Debian and Red Hat-based distros.

sauterp:
In the discussion of #492 it became apparent that some users are lacking an easy and automatic way to install the latest version of exo cli. We add a script to automate this for Debian and Red Hat-based distributions.

The script requires `sh`, `sha256sum`(to check the checksum), `curl` or `wget` and either `dpkg`, `yum` or `dnf`. It detects automatically which one of these commands should be used to download and install the package.

The script can be run in one command with `curl -fsSL <url to script> | sh` and apart from invoking `sudo` or `doas` it should not require any more user interaction.


kobajagi (approved):

exo-cedric (approved):
:smile: 

(one nice addition would be to sign and verify the `checksums.txt`)

PR #515, opened 2023-05-31
Ignore regex checks in `validateDatabaseServiceSettings`

kobajagi:
Fixes the issue with validation error for grafana settings `smtp.from_address`. It was identified as a regex pattern originating from Aiven that does not match in Go. Because of this we cannot rely on the clients side validation of patterns and I've chosen to disable the feature and rely on server side validation.

```
$ go run . dbaas update mydb --grafana-settings '{"smtp_server":{"host":"example.net","port":587,"username":"person@example.net","password":"mypassword","from_address":"dashboard@example.net"}}'                                      
 ✔ Updating Database Service "mydb"... 0s                                                                                    
```

### DIFF
--- a/cmd/dbaas.go
+++ b/cmd/dbaas.go
@@ -78,16 +78,18 @@ func validateDatabaseServiceSettings(in string, schema interface{}) (map[string]
 	}
 
 	if !res.Valid() {
-		return nil, errors.New(strings.Join(
-			func() []string {
-				errs := make([]string, len(res.Errors()))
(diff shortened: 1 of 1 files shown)


PhilippeChepy (approved):

sauterp (approved):
it's a good solution.
If we ever really need to match regex patterns that work in other languages, then I think we need to use a regex package that implements PCRE2, AFAIK that's what python uses. But that would be overkill here.

PR #518, opened 2023-06-08
storage list: allow listing versions of objects

sauterp:
We add the `--versions` flag to the `exo storage list` command. If enabled all versions of objects in the bucket will be listed. The logic for object listing was extended with a filtering mechanism that allows supplying any number of filter functions for both versioned and non-versioned objects. Flags for filtering objects by age were added.
Later we will be able to reuse the same filtering logic in all `storage` sub-commands that operate on lists of objects.


PhilippeChepy (approved):
:+1:

kobajagi (approved):

PR #521, opened 2023-06-21
Sauterp/60749/in and exclude versions flags

sauterp:
We implement the flags `--exclude-versions` and `--only-versions` which are mutually exclusive and allow the user to limit the versions listed to a set of version numbers and version IDs. We also add the flag `--exclude-latest-version`.

The versions of objects are sorted by last modified timestamps and then version numbers starting with `v0` are assigned to the oldest version and `v1` etc. to later versions. This means version numbers are


PhilippeChepy (approved):

pkg/flags/version_filters.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return err
+	}
+
+	vsToExclude, err := cmd.Flags().GetStringSlice(ExcludeVersions)
+	if err != nil {
+		return err
+	}
+
+	if len(vsToInclude) > 0 && len(vsToExclude) > 0 {
```
kobajagi, line 58: Isn't `ExcludeLatestVersion` also mutually exclusive with `OnlyVersions`?

kobajagi (approved):

PR #522, opened 2023-07-06
create release GH Action workflow

sauterp:
We create a GH Action release workflow that builds, tests and releases exo cli using the release Action from go.mk.


.github/workflows/release.yml
```diff
@@ -0,0 +1,23 @@
+name: release
+
+on:
+  push:
+    tags:
```
kobajagi, line 5: Perhaps we should also specify master branch constraint? Just in case we accidentally tag a branch commit.

kobajagi (changes requested):

.github/workflows/release.yml
sauterp: I tried that, unfortunately it turns out that if you set multiple triggers, they are matched with a logical 'OR'. This is not explicitly documented, but I verified it on my test repo.

kobajagi: Well that is no fun :smirk:

kobajagi (approved):

PhilippeChepy (approved):

PR #619, opened 2024-07-12
Compute Instance delete: Remove multiple entities by their IDs/Names

pierre-emmanuelJ:
# Description
Compute Instance delete, remove multiple entities by their IDs/Names

## Checklist
(For exoscale contributors)

* [x] Testing

## Testing

Deleting a list of instance
```BASH
go run main.go c i delete test1 test2 test3
[+] Are you sure you want to delete instance "test1"? [yN]: y
[+] Are you sure you want to delete instance "test2"? [yN]: y
[+] Are you sure you want to delete instance "test3"? [yN]: y
 ✔ Deleting instance "test1, test2, test3"... 9s
```

Not existing instance in list
```
go run main.go c i delete test1 test2 test404 test3
[+] Are you sure you want to delete instance "test1"? [yN]: y
[+] Are you sure you want to delete instance "test2"? [yN]: y
error: "test404" not found in ListInstancesResponse: resource not found
exit status 1
```

Force deletion like rm cmd even if not exist
```
go run main.go c i delete test1 test2 test404 test3 testNotexist -f
 ✔ Deleting instance "test1, test2, test404, test3, testNotexist"... 9s
```

Nothing happen here
```
go run main.go c i delete 404 dummy notfound -f
```

### DIFF
--- a/cmd/instance_delete.go
+++ b/cmd/instance_delete.go
@@ -54,10 +54,10 @@ func (c *instanceDeleteCmd) cmdRun(_ *cobra.Command, _ []string) error {
 	for _, i := range c.Instances {
 		instance, err := instances.FindListInstancesResponseInstances(i)
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "warning: %s not found.\n", i)
 			if !c.Force {
 				return err
 			}
+			fmt.Fprintf(os.Stderr, "warning: %s not found.\n", i)
 
 			continue
 		}
@@ -88,6 +88,8 @@ func (c *instanceDeleteCmd) cmdRun(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
+	// Cleaning up resources created in create instance
+	// https://github.com/exoscale/cli/blob/master/cmd/instance_create.go#L220
 	for _, i := range instanceToDelete {
 		instanceDir := path.Join(globalstate.ConfigFolder, "instances", i.String())
 		if _, err := os.Stat(instanceDir); !os.IsNotExist(err) {
--- a/cmd/instance_delete.go
+++ b/cmd/instance_delete.go
@@ -1,16 +1,15 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"path"
+	"strings"
 
 	"github.com/spf13/cobra"
 
-	"github.com/exoscale/cli/pkg/account"
 	"github.com/exoscale/cli/pkg/globalstate"
-	exoapi "github.com/exoscale/egoscale/v2/api"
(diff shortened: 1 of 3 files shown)


cmd/instance_delete.go
```diff
@@ -36,33 +35,64 @@ func (c *instanceDeleteCmd) cmdPreRun(cmd *cobra.Command, args []string) error {
...
-	instanceDir := path.Join(globalstate.ConfigFolder, "instances", *instance.ID)
-	if _, err := os.Stat(instanceDir); !os.IsNotExist(err) {
-		if err := os.RemoveAll(instanceDir); err != nil {
-			return fmt.Errorf("error deleting instance directory: %w", err)
+	for _, i := range instanceToDelete {
+		instanceDir := path.Join(globalstate.ConfigFolder, "instances", i.String())
+		if _, err := os.Stat(instanceDir); !os.IsNotExist(err) {
+			if err := os.RemoveAll(instanceDir); err != nil {
+				return fmt.Errorf("error deleting instance directory: %w", err)
+			}
```
sauterp, line 95: What does this section do? Might we worth a commend.

sauterp (approved):

cmd/instance_delete.go
pierre-emmanuelJ: There is a mechanism at instance creation to create default generated sshkeys (private key stored in dedicated dir and public key registered to the instance)
https://github.com/exoscale/cli/blob/master/cmd/instance_create.go#L220

That let you ssh to it if you did not provide a sshkey.

The big benefit of that, is you can directly use the command `exo compute instance ssh <instance name/id>` and it will do the magic for you

sauterp: and the relation to this section is that it cleans up the generated keys, I suppose.
you could write a comment in the code with this info.

cmd/instance_delete.go
```diff
@@ -36,33 +35,64 @@ func (c *instanceDeleteCmd) cmdPreRun(cmd *cobra.Command, args []string) error {
...
-		if !askQuestion(fmt.Sprintf("Are you sure you want to delete instance %q?", c.Instance)) {
-			return nil
+	instanceToDelete := []v3.UUID{}
+	for _, i := range c.Instances {
+		instance, err := instances.FindListInstancesResponseInstances(i)
+		if err != nil {
+			if !c.Force {
+				return err
+			}
+
```
kobajagi, line 60: I think it would be useful to print error (to `stderr`) before `continue`. Flag `--force` is to not show confirmation dialog, but it doesn't have to be silent command and we can still print error.

kobajagi (approved):

PR #627, opened 2024-08-05
storage: handle errors in batch objects delete action

kobajagi:
# Description

Batch objects delete API method from `aws-sdk-go-v2` will not return error as second argument if individual objects failed to be deleted. Instead it will return both list of deleted objects and list of errors in result structure.

This prints errors (if any) after the list of successfully deleted files.

## Checklist
(For exoscale contributors)

* [ ] Testing

### DIFF
--- a/pkg/storage/sos/object.go
+++ b/pkg/storage/sos/object.go
@@ -31,6 +31,28 @@ import (
 	"github.com/exoscale/cli/utils"
 )
 
+// BatchErrorList contains list of errors parsed from AWS SDK types.DeletedObject Error attribute.
+// Only Message attribute of each types.Error object is taken and converted to std Error.
+type BatchErrorList struct {
+	List []error
+}
+
+func NewBatchErrorList() *BatchErrorList {
+	return &BatchErrorList{
+		[]error{},
+	}
+}
+
+func (l BatchErrorList) Error() string {
+	return fmt.Sprintf("batch job returned %d errors", len(l.List))
+}
+
+func (l *BatchErrorList) AddErrors(errs []types.Error) {
+	for _, err := range errs {
+		l.List = append(l.List, fmt.Errorf("%s", err.Message))
+	}
+}
+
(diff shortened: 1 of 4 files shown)


pkg/storage/sos/object.go
```diff
@@ -31,6 +31,28 @@ import (
 	"github.com/exoscale/cli/utils"
 )
 
+// BatchErrorList contains list of errors parsed from AWS SDK types.DeletedObject Error attribute.
+// Only Message attribute of each types.Error object is taken and converted to std Error.
+type BatchErrorList struct {
+	List []error
+}
+
+func NewBatchErrorList() *BatchErrorList {
```
pierre-emmanuelJ, line 40: you can probably reuse the already imported package in the CLI:
https://github.com/exoscale/cli/blob/master/cmd/output.go#L8

pkg/storage/sos/object.go
```diff
@@ -61,9 +63,21 @@ func (c *Client) DeleteObjects(ctx context.Context, bucket, prefix string, recur
 		}
 
 		deleted = append(deleted, res.Deleted...)
+		for _, err := range res.Errors {
+			var e error
+			switch {
+			case err.Message != nil:
+				e = errors.New(*err.Message)
+			case err.Code != nil:
+				e = errors.New(*err.Code)
```
sauterp, line 72: Not sure it's a problem, but if it's possible for an error to have a Message AND a Code, this switch will only add the Message AFAIU and the Code would be ignored.

sauterp (approved):

pkg/storage/sos/object.go
kobajagi: Errors are expected to have both Message & Code. Code is intended for machine parsing where Message is human readable error description. We only print errors (we won't parse Code), this is simple fallback in case of exception in SOS so we have some error printed.

kobajagi: If we later decide to implement control flow.around Code, we can introduce custom Error. For now this should be enough.

pierre-emmanuelJ (approved):
😃

PR #628, opened 2024-08-14
Fix list of dependencies for archlinux builds

kobajagi:
# Description

For archlinux builds install `base-devel` metapackage instead of individual tools. This fixes the build error triggered by missing `debugedit` tool:

```
==> ERROR: Cannot find the debugedit binary required for including source files in debug packages.
```

`base-devel` metapackage contains following packages (as the time of writing):
```
$ pacman -Qi base-devel | grep Depends
Depends On      : archlinux-keyring  autoconf  automake  binutils  bison  debugedit  fakeroot  file  findutils  flex  gawk  gcc  gettext  grep  groff  gzip  libtool  m4  make  pacman  patch  pkgconf  sed  sudo  texinfo  which
```

## Checklist
(For exoscale contributors)

* [ ] Testing

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,7 +71,7 @@ jobs:
         shell: bash
 
       - name: install tools
-        run: pacman --noconfirm -Sy git openssh glibc sudo binutils make gcc pkg-config fakeroot go which
+        run: pacman --noconfirm -Sy base-devel git openssh glibc go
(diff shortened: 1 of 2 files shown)


rokernel (approved):

sauterp (approved):
thanks!

7fELF (approved):

PR #640, opened 2024-09-21
refactor(dns): Update dns manipulation to egoscale v3

mlec1:
# Description
Update dns manipulation to egoscale v3

## Checklist
(For exoscale contributors)

* [ ] Testing

## Testing

Manual test of the different dns manipulation.

### DIFF
--- a/cmd/dns.go
+++ b/cmd/dns.go
@@ -6,10 +6,8 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/exoscale/cli/pkg/account"
 	"github.com/exoscale/cli/pkg/globalstate"
-	exo "github.com/exoscale/egoscale/v2"
-	exoapi "github.com/exoscale/egoscale/v2/api"
+	v3 "github.com/exoscale/egoscale/v3"
 )
 
 var dnsCmd = &cobra.Command{
@@ -18,19 +16,28 @@ var dnsCmd = &cobra.Command{
 }
 
 // domainFromIdent returns a DNS domain from identifier (domain name or ID).
-func domainFromIdent(ident string) (*exo.DNSDomain, error) {
-	ctx := exoapi.WithEndpoint(gContext, exoapi.NewReqEndpoint(account.CurrentAccount.Environment, account.CurrentAccount.DefaultZone))
-	if exo.IsValidUUID(ident) {
-		return globalstate.EgoscaleClient.GetDNSDomain(ctx, account.CurrentAccount.DefaultZone, ident)
+func domainFromIdent(ident string) (*v3.DNSDomain, error) {
+	ctx := gContext
+
+	domainId, err := v3.ParseUUID(ident)
+	if err == nil {
+		// ident is a valid UUID
+		domain, err := globalstate.EgoscaleV3Client.GetDNSDomain(ctx, domainId)
+		if err != nil {
+			return nil, err
+		}
+
+		return domain, nil
 	}
 
-	domains, err := globalstate.EgoscaleClient.ListDNSDomains(ctx, account.CurrentAccount.DefaultZone)
+	// ident is not a UUID, trying finding domain by name
+	domains, err := globalstate.EgoscaleV3Client.ListDNSDomains(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	for _, domain := range domains {
-		if *domain.UnicodeName == ident {
+	for _, domain := range domains.DNSDomains {
+		if domain.UnicodeName == ident {
 			return &domain, nil
 		}
 	}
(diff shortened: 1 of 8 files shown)


cmd/dns.go
```diff
@@ -18,59 +15,41 @@ var dnsCmd = &cobra.Command{
...
+	domainsResp, err := globalstate.EgoscaleV3Client.ListDNSDomains(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	for _, domain := range domains {
-		if *domain.UnicodeName == ident {
-			return &domain, nil
-		}
+	domain, err := domainsResp.FindDNSDomain(ident)
```
pierre-emmanuelJ, line 26: 👍

cmd/dns_add.go
```diff
@@ -20,28 +19,79 @@ func init() {
...
-		_, err = globalstate.EgoscaleClient.CreateDNSDomainRecord(ctx, account.CurrentAccount.DefaultZone, *domain.ID, &exo.DNSDomainRecord{
-			Name:     &name,
-			Type:     &rType,
-			Content:  &content,
-			TTL:      &ttl,
-			Priority: priority,
-		})
+	ctx := gContext
+	err = decorateAsyncOperations(fmt.Sprintf("Adding DNS record %q to %q...", rType, domain.UnicodeName), func() error {
+
```
pierre-emmanuelJ, line 58:
```suggestion
```

cmd/dns_list.go
```diff
@@ -57,18 +54,23 @@ Supported output template annotations: %s`,
...
+
+		// Convert v3.UUID to string using String() method, then get a pointer to it
+		// Don't know if it is best practice
+		idStr := d.ID.String() // Convert UUID to string
+
 		o := dnsListItemOutput{
-			ID:   StrPtrFormatOutput(d.ID),
-			Name: StrPtrFormatOutput(d.UnicodeName),
+			ID:   StrPtrFormatOutput(&idStr),
+			Name: StrPtrFormatOutput(&d.UnicodeName),
```
pierre-emmanuelJ, line 73: I think here you can simplify by removing pointer from `dnsListItemOutput`

cmd/dns_show.go
```diff
@@ -70,53 +68,53 @@ func showDNS(ident, name string, types []string) (output.Outputter, error) {
...
 		}
 
 		var ttl int64
-		if record.TTL != nil {
-			ttl = *record.TTL
+		if record.Ttl != 0 {
+			ttl = record.Ttl
 		}
 
 		out = append(out, dnsShowItemOutput{
```
pierre-emmanuelJ, line 108: same here probably `dnsShowItemOutput` can be uopdated

pierre-emmanuelJ:
This PR looks very good, :) I will do a second review once you applied some of them

cmd/dns.go
```diff
@@ -18,59 +15,41 @@ var dnsCmd = &cobra.Command{
...
+	if err != nil {
+		return nil, err
 	}
 
-	return nil, fmt.Errorf("domain %q not found", ident)
+	return &domain, err
```
sauterp, line 31:
```suggestion
	return &domain, nil
```

cmd/dns_add.go
```diff
@@ -20,28 +19,79 @@ func init() {
...
+	"CNAME": v3.CreateDNSDomainRecordRequestTypeCNAME,
+	"SSHFP": v3.CreateDNSDomainRecordRequestTypeSSHFP,
+	"SRV":   v3.CreateDNSDomainRecordRequestTypeSRV,
+	"AAAA":  v3.CreateDNSDomainRecordRequestTypeAAAA,
+	"MX":    v3.CreateDNSDomainRecordRequestTypeMX,
+	"TXT":   v3.CreateDNSDomainRecordRequestTypeTXT,
+	"ALIAS": v3.CreateDNSDomainRecordRequestTypeALIAS,
+	"URL":   v3.CreateDNSDomainRecordRequestTypeURL,
+	"SPF":   v3.CreateDNSDomainRecordRequestTypeSPF,
+}
```
sauterp, line 39:
```
var dnsRecordTypeMap = map[string]v3.CreateDNSDomainRecordRequestType{
	string(v3.CreateDNSDomainRecordRequestTypeNS): v3.CreateDNSDomainRecordRequestTypeNS,
	// and so on for each type
}
```
you could do this instead, leaves less room for spelling errors.

sauterp (commented):
thanks!

cmd/dns_add.go
pierre-emmanuelJ: We can probably use stringer here in egoscale what do you think @sauterp ?

cmd/dns_add.go
```diff
@@ -55,8 +55,7 @@ func addDomainRecord(domainIdent, name, rType, content string, ttl int64, priori
 
 	ctx := gContext
 	err = decorateAsyncOperations(fmt.Sprintf("Adding DNS record %q to %q...", rType, domain.UnicodeName), func() error {
-
-		recordType, err := StringToDNSDomainRecordRequestType(rType)
+		recordType := v3.CreateDNSDomainRecordRequestType("TEST")
```
sauterp, line 58: looks like you forgot to change this?

cmd/dns_remove.go
```diff
@@ -54,11 +54,7 @@ func removeDomainRecord(domainIdent, recordIdent string, force bool) error {
 		}
 
 		_, err = globalstate.EgoscaleV3Client.Wait(ctx, op, v3.OperationStateSuccess)
-		if err != nil {
-			return fmt.Errorf("exoscale: error while waiting DNS record deletion: %w", err)
```
sauterp: You're right that these can be removed, but I like these error messages. Makes it easier to find out where an error happened when you read the logs. IMHO these can be kept :)

cmd/dns_add.go
```diff
@@ -20,29 +19,72 @@ func init() {
...
+	string(v3.CreateDNSDomainRecordRequestTypeAAAA):  v3.CreateDNSDomainRecordRequestTypeAAAA,
+	string(v3.CreateDNSDomainRecordRequestTypeMX):    v3.CreateDNSDomainRecordRequestTypeMX,
+	string(v3.CreateDNSDomainRecordRequestTypeTXT):   v3.CreateDNSDomainRecordRequestTypeTXT,
+	string(v3.CreateDNSDomainRecordRequestTypeALIAS): v3.CreateDNSDomainRecordRequestTypeALIAS,
+	string(v3.CreateDNSDomainRecordRequestTypeURL):   v3.CreateDNSDomainRecordRequestTypeURL,
+	string(v3.CreateDNSDomainRecordRequestTypeSPF):   v3.CreateDNSDomainRecordRequestTypeSPF,
+}
+
+// Function to get the DNSDomainRecordRequestType from a string
+func StringToDNSDomainRecordRequestType(recordType string) (v3.CreateDNSDomainRecordRequestType, error) {
```
pierre-emmanuelJ, line 42:
```suggestion
// StringToDNSDomainRecordRequestType gets the DNSDomainRecordRequestType from a string
func StringToDNSDomainRecordRequestType(recordType string) (v3.CreateDNSDomainRecordRequestType, error) {
```

cmd/dns_add.go
```diff
@@ -20,29 +19,72 @@ func init() {
...
+	// Lookup the record type in the map
```
pierre-emmanuelJ, line 43:
```suggestion
```
IMHO code is enough explicit here

cmd/dns_remove.go
pierre-emmanuelJ: @mlec1 What I wanted to point out here in this PR, if you were not wrapping/modifying the error, it was not needed to add an extra condition and return it directly instead.
But in the case you want to add more context on the error like here, that is fine 👍 you can keep it.

I have no strong opinion here, but if you decided to add more context to the error here, looks good to me.

PR #642, opened 2024-10-06
refactor(eip): Refactor eip using egoscale v3

mlec1:
Update the elastic ip using Egoscale v3. 
Some stuff are not working yet: I put some comments in the code, you will see.

## Checklist
(For exoscale contributors)

* [ ] Testing

## Testing

Create, delete, update ip,...

### DIFF
--- a/cmd/elastic_ip_create.go
+++ b/cmd/elastic_ip_create.go
@@ -94,13 +94,8 @@ func (c *elasticIPCreateCmd) cmdRun(_ *cobra.Command, _ []string) error {
 		}
 
 		_, err = client.Wait(ctx, op, v3.OperationStateSuccess)
-		if err != nil {
-			return fmt.Errorf("exoscale: error while waiting for Elastic IP creation: %w", err)
-		}
-
-		return nil
+		return err
 	})
-
 	if err != nil {
 		return err
 	}
--- a/cmd/elastic_ip_create.go
+++ b/cmd/elastic_ip_create.go
@@ -94,7 +94,11 @@ func (c *elasticIPCreateCmd) cmdRun(_ *cobra.Command, _ []string) error {
 		}
 
 		_, err = client.Wait(ctx, op, v3.OperationStateSuccess)
-		return err
+		if err != nil {
+			return fmt.Errorf("exoscale: error while waiting for Elastic IP creation: %w", err)
+		}
+
+		return nil
 	})
 	if err != nil {
 		return err
--- a/cmd/elastic_ip_create.go
+++ b/cmd/elastic_ip_create.go
@@ -3,16 +3,12 @@ package cmd
 import (
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/spf13/cobra"
 
-	"github.com/exoscale/cli/pkg/account"
 	"github.com/exoscale/cli/pkg/globalstate"
 	"github.com/exoscale/cli/pkg/output"
-	"github.com/exoscale/cli/utils"
-	egoscale "github.com/exoscale/egoscale/v2"
-	exoapi "github.com/exoscale/egoscale/v2/api"
+	v3 "github.com/exoscale/egoscale/v3"
 )
 
 type elasticIPCreateCmd struct {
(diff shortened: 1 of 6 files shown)


cmd/elastic_ip_create.go
```diff
@@ -53,56 +49,70 @@ func (c *elasticIPCreateCmd) cmdPreRun(cmd *cobra.Command, args []string) error
 }
 
 func (c *elasticIPCreateCmd) cmdRun(_ *cobra.Command, _ []string) error {
-	ctx := exoapi.WithEndpoint(gContext, exoapi.NewReqEndpoint(account.CurrentAccount.Environment, c.Zone))
+	ctx := gContext
+	client, err := switchClientZoneV3(ctx, globalstate.EgoscaleV3Client, c.Zone)
```
pierre-emmanuelJ, line 53: 👍

cmd/elastic_ip_create.go
```diff
@@ -53,56 +49,70 @@ func (c *elasticIPCreateCmd) cmdPreRun(cmd *cobra.Command, args []string) error
...
+		}
+
+		_, err = client.Wait(ctx, op, v3.OperationStateSuccess)
+		if err != nil {
+			return fmt.Errorf("exoscale: error while waiting for Elastic IP creation: %w", err)
+		}
+
+		return nil
 	})
+
```
pierre-emmanuelJ, line 103:
```suggestion
```

cmd/elastic_ip_create.go
```diff
@@ -53,56 +49,70 @@ func (c *elasticIPCreateCmd) cmdPreRun(cmd *cobra.Command, args []string) error
...
 	return (&elasticIPShowCmd{
 		cliCommandSettings: c.cliCommandSettings,
-		ElasticIP:          *elasticIP.ID,
-		Zone:               c.Zone,
+		// Comment for reviewer:
+		// Is there a way to get the created Elastic IP address or UUID ???
+		// Listing all of them and comparing Addressfamily, Description,... doesn't really garantee uniquess
+		// TODO: Remove comment before merging
+		ElasticIP: "",
+		Zone:      c.Zone,
```
pierre-emmanuelJ, line 115: yes you need to get the `op` after waiting and there is the `op.Reference.ID` check if reference not nil before using

cmd/elastic_ip_delete.go
```diff
@@ -52,9 +60,20 @@ func (c *elasticIPDeleteCmd) cmdRun(_ *cobra.Command, _ []string) error {
...
+		}
+
+		_, err = client.Wait(ctx, op, v3.OperationStateSuccess)
+		if err != nil {
+			return fmt.Errorf("exoscale: error while waiting for Elastic IP deletion: %w", err)
+		}
+
+		return nil
 	})
+
```
pierre-emmanuelJ, line 76:
```suggestion
```

cmd/elastic_ip_list.go
```diff
@@ -50,51 +49,47 @@ func (c *elasticIPListCmd) cmdPreRun(cmd *cobra.Command, args []string) error {
...
-			res <- elasticIPListItemOutput{
-				ID:        *e.ID,
-				IPAddress: e.IPAddress.String(),
-				Zone:      zone,
-			}
+		for _, elasticIP := range resp.ElasticIPS {
+			output = append(output, elasticIPListItemOutput{
+				ID:          elasticIP.ID,
+				IPAddress:   elasticIP.IP,
+				Zone:        string(zone.Name),
```
pierre-emmanuelJ, line 86: no need the `string()` if you updated the type in output, this such change apply to this PR and other one :)

utils/utils.go
```diff
@@ -76,8 +76,14 @@ func GetInstancesInSecurityGroup(ctx context.Context, client *v2.Client, securit
...
-func GetInstancesAttachedToEIP(ctx context.Context, client *v2.Client, elasticIPID, zone string) ([]*v2.Instance, error) {
-	instances, err := client.ListInstances(ctx, zone, v2.ListInstancesByIpAddress(elasticIPID))
+func GetInstancesAttachedToEIP(ctx context.Context, client *v3.Client, elasticIPID string) (*v3.ListInstancesResponse, error) {
+	// Message for reviewer:
+	// It always returns the list of all instances. Why ???
+	// When no instance attached to Elastic IP, should returned an empty list
+	// When some instances attached to elastic IP, should returns only the attached instances
+	// Am I doing something wrong here ?? Is the ip-address parameter taking the Elastic IP into account ??
+	// TODO: remove comment before merging
+	instances, err := client.ListInstances(ctx, v3.ListInstancesWithIPAddress(elasticIPID))
```
pierre-emmanuelJ, line 86: I'll try to reproduce, thanks for reporting

cmd/elastic_ip_show.go
```diff
@@ -106,58 +104,66 @@ func (c *elasticIPShowCmd) cmdPreRun(cmd *cobra.Command, args []string) error {
...
-		out.HealthcheckURI = elasticIP.Healthcheck.URI
-		out.HealthcheckInterval = elasticIP.Healthcheck.Interval
-		out.HealthcheckTimeout = elasticIP.Healthcheck.Timeout
-		out.HealthcheckStrikesOK = elasticIP.Healthcheck.StrikesOK
-		out.HealthcheckStrikesFail = elasticIP.Healthcheck.StrikesFail
-		out.HealthcheckTLSSNI = elasticIP.Healthcheck.TLSSNI
-		out.HealthcheckTLSSkipVerify = elasticIP.Healthcheck.TLSSkipVerify
+		// Message for reviewer:
+		// I don't really like to use this function for only one this parameter, if you have a better solution, please feel free to suggest it
+		// TODO: remove comment before merging
```
pierre-emmanuelJ, line 157: This part is not gorgeous I agree, it's hard to define what's an empty value for certain type if we are not relying on pointers, example for int or bool, probably for sting types, you can remove pointer and rely only omitempty tag fpr empty string.
Probably need to be verified for other types like int, if in this case we can consider 0 as an empty value...etc

pierre-emmanuelJ (commented):
It's a good change!!, we will do another review on it

cmd/elastic_ip_delete.go
```diff
@@ -60,25 +60,15 @@ func (c *elasticIPDeleteCmd) cmdRun(_ *cobra.Command, _ []string) error {
...
-	err = decorateAsyncOperations(fmt.Sprintf("Deleting Elastic IP %s...", c.ElasticIP), func() error {
+	return decorateAsyncOperations(fmt.Sprintf("Deleting Elastic IP %s...", c.ElasticIP), func() error {
 		op, err := client.DeleteElasticIP(ctx, elasticIP.ID)
 		if err != nil {
 			return fmt.Errorf("exoscale: error while deleting Elastic IP: %w", err)
 		}
 
 		_, err = client.Wait(ctx, op, v3.OperationStateSuccess)
-		if err != nil {
-			return fmt.Errorf("exoscale: error while waiting for Elastic IP deletion: %w", err)
```
pierre-emmanuelJ: As I pointed out in another PR, if you decided to add more context on the error you can keep, it, I was pointing the change with no context added.

To stay consistent, everywhere, it's better if you not do a mix of sometime add error wrapping and sometime not, you decide, you wrap all and return the proper error, or you wrap not.

IMO, having error wrapping with more context is always better in case of error :) but make sure to respect the same behavior everywhere and stay consistent

pierre-emmanuelJ (changes requested):

utils/utils.go
mlec1: The list-instance option is still not working. 
My guess is that in `egoscale`, the get parameter being added is `ipAddress` https://github.com/exoscale/egoscale/blob/v3.1.7/v3/operations.go#L9770 when on the [openapi](https://openapi-v2.exoscale.com/operation/operation-list-instances#operation-list-instances-query) spec the query parameter should be `ip-address`.

Same for `manager-id` and `manager-type`

It is just a supposition.

cmd/elastic_ip_create.go
mlec1: Is there a way to get the `op` from the `decorateAsyncOperations` ?? 
As this function is only returning an error if it happens ??!!

tgrondier:
Replaced by https://github.com/exoscale/cli/pull/684

PR #654, opened 2024-11-29
DBAAS: support for user operations

tgrondier:
# Description
<!--
* Prefix: the title with the component name being changed. Add a short and self describing sentence to ease the review
* Please add a few lines providing context and describing the change
* Please self comment changes whenever applicable to help with the review process
* Please keep the checklist as part of the PR. Tick what applies to this change.
-->

This adds for the following API calls:

- Creating users for supported DBAAS services
- Deleting users for supported DBAAS services
- Resetting user credentials for supported DBAAS services
- Listing and showing user details for supported DBAAS services


## Checklist
(For exoscale contributors)

* [x] Testing

## Testing

```





➜  ~/exo/cli git:(tgrondier/sc-104242/cli-users) ✗ go run . dbaas user delete test myuser3                                                                                                                                  24-11-29 16:12 
[+] Are you sure you want to delete user "myuser3" [yN]: y
 ✔ Creating DBaaS user "myuser3" 0s
┼───────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼
│   DATABASE SERVICE    │                                                                                                                              │
┼───────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼
│ Zone                  │ ch-gva-2                                                                                                                     │
│ Name                  │ test                                                                                                                         │
│ Type                  │ mysql                                                                                                                        │
│ Plan                  │ hobbyist-2                                                                                                                   │
│ Disk Size             │ 8.0 GiB                                                                                                                      │
│ State                 │ running                                                                                                                      │
│ Creation Date         │ 2024-11-29 15:00:34 +0000 UTC                                                                                                │
│ Update Date           │ 2024-11-29 15:02:56 +0000 UTC                                                                                                │
│ Nodes                 │ 1                                                                                                                            │
│ Node CPUs             │ 2                                                                                                                            │
│ Node Memory           │ 2.0 GiB                                                                                                                      │
│ Termination Protected │ false                                                                                                                        │
│ Maintenance           │ saturday (19:26:58)                                                                                                          │
...
│ Users                 │ avnadmin (primary)                                                                                                           │

┼───────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼



➜  ~/exo/cli git:(tgrondier/sc-104242/cli-users) ✗ go run . dbaas user reset-credentials test myuser2                                                                                                                                                                                                       24-12-02 10:52 
 ✔ Resetting DBaaS user "myuser2" 0s
┼────────────────┼──────────────────────────┼
│  SERVICE USER  │                          │
┼────────────────┼──────────────────────────┼
│ Username       │ myuser2                  │
│ Type           │ regular                  │
│ Password       │ AVNS_46o-1gO9pE8acEc9Nzp │
│ Access Control │   Categories             │
│                │   Channels     *         │
│                │   Commands               │
│                │   Keys                   │
│                │                          │
┼────────────────┼──────────────────────────┼


➜  ~/exo/cli git:(tgrondier/sc-104242/cli-users) ✗ go run . dbaas user create test myuser2                                                                                                                                                                                                                  24-12-02 10:46 
 ✔ Creating DBaaS user "myuser2" 0s
┼────────────────┼──────────────────────────┼
│  SERVICE USER  │                          │
┼────────────────┼──────────────────────────┼
│ Username       │ myuser2                  │
│ Type           │ normal                   │
│ Password       │ AVNS_QWLuqR0A_FCAll3JiVN │
│ Authentication │ caching_sha2_password    │
┼────────────────┼──────────────────────────┼

```


clovisgenevard:
Hello! Could someone provide some context about this release, the roadmap, or anything else relevant so @ridge can work on the changelog and community update?

cmd/dbaas_user_show_redis.go
```diff
@@ -0,0 +1,70 @@
+package cmd
```
ste-m, line 1: Let's not add Redis.
Aiven is deprecating it, and it will be soon to be removed from our CLI as well.

cmd/dbaas_user_reset_redis.go
```diff
@@ -0,0 +1,49 @@
+package cmd
```
ste-m, line 1: Same here no Redis

cmd/dbaas_user_show_redis.go
tgrondier: Removed in efdee4f2

tgrondier:
Removed all redis support from this PR ( https://github.com/exoscale/cli/pull/654/commits/efdee4f2190dc3c5f5430974df46af52e634e28f )

ste-m:
> Hello! Could someone provide some context about this release, the roadmap, or anything else relevant so @ridge can work on the changelog and community update?

Nothing super new @clovisgenevard , we are simply aligning the clients.
When we release a feature we reach API level, then clients follow on their own rhythm.
Sometimes some parts are being left behind for a while. In this case the features exists since long, but the CLI had no opportunity to follow.
If you want to communicate you can do something on the lines of "DBaaS user management is now available in the CLI".

cmd/dbaas_user_show_kafka.go
```diff
@@ -0,0 +1,56 @@
...
+	s, err := client.GetDBAASServiceKafka(ctx, c.Name)
+	if err != nil {
+		return &dbaasUserShowOutput{}, err
+	}
+
+	for _, u := range s.Users {
+		if u.Username == c.Username {
+			return &dbaasUserShowOutput{
+				Username: c.Username,
+				Password: u.Password,
```
ste-m, line 43: I don't think this should be here. In general all secrets should not be returned by the "show SERVICE user" commands. While as per today you might find the information you want in them, they will be removed, and you would end up showing empty strings. There are specific reveal-secret commands afaik.
Or maybe i did misunderstand, but in this case maybe we could review together in a call?

tgrondier: Your understanding is correct, this was indeed the password as returned by Aiven's API and it was shown as is. Removed from all `show` commands.

tgrondier: Added a command `reveal-secrets` (3e80a064) specifically to display passwords and other secrets

ste-m:
> redis is going to be replaced as a drop in by valkey under the same naming: Caching, so not sure we should avoid doing redis now.

For one thing, the commands will all change name. To my understanding it's NOT an in-place substitution, is a NEW service, and customer will need to adapt to new endpoints, new commands, new naming etc etc. Their services will not just renamed "Valkey" under the hood, they will be MIGRATED to Valkey. I would not encourage using redis when we are going to deprecate it by the end of the month: it will soon not be possible anymore to create a new redis services and i doubt we will have Valkey before Februray at soonest.

pierre-emmanuelJ (approved):
Great work @tgrondier !

ste-m (approved):

kobajagi (approved):

PR #685, opened 2025-05-19
IAM org policy and roles: move to egoscale v3

tgrondier:
# Description

Move all remaining IAM commands to use exclusively egoscale v3

## Checklist
(For exoscale contributors)

* [x] Testing

## Testing

Testing logs in individual commit messages

### DIFF
--- a/go
+++ b/go
@@ -1,0 +1,9 @@
+{
+  default-service-strategy: allow,
+  services: {
+    iam: {
+      type: allow
+    }
+  }
+}
+ run . iam org-policy update -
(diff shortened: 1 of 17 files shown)


go
```diff
@@ -0,0 +1,9 @@
+{
+  default-service-strategy: allow,
+  services: {
+    iam: {
+      type: allow
+    }
+  }
+}
+ run . iam org-policy update -
```
sauterp, line 9: did you commit this file by accident?

tgrondier: Fixed in 128baae8, looks it was the result of an incorrect pipe

pierre-emmanuelJ (approved):

PR #687, opened 2025-05-20
load-balancer: move to egoscale v3

tgrondier:
# Description

Move load-balancer commands to use exclusively egoscale v3

## Checklist
(For exoscale contributors)

* [x] Testing

## Testing

In individual commit messages

### DIFF
--- a/cmd/nlb_create.go
+++ b/cmd/nlb_create.go
@@ -20,7 +20,7 @@ type nlbCreateCmd struct {
 
 	Description string            `cli-usage:"Network Load Balancer description"`
 	Labels      map[string]string `cli-flag:"label" cli-usage:"Network Load Balancer label (format: key=value)"`
-	Zone        string            `cli-short:"z" cli-usage:"Network Load Balancer zone"`
+	Zone        v3.ZoneName       `cli-short:"z" cli-usage:"Network Load Balancer zone"`
 }
 
 func (c *nlbCreateCmd) cmdAliases() []string { return gCreateAlias }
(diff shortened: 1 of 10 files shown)


cmd/nlb_create.go
```diff
@@ -43,22 +40,26 @@ func (c *nlbCreateCmd) cmdPreRun(cmd *cobra.Command, args []string) error {
...
+	nlb := v3.CreateLoadBalancerRequest{
+		Description: c.Description,
+		Labels:      c.Labels,
+		Name:        c.Name,
 	}
+	var err error
 
-	ctx := exoapi.WithEndpoint(gContext, exoapi.NewReqEndpoint(account.CurrentAccount.Environment, c.Zone))
+	ctx := gContext
+	client, err := switchClientZoneV3(ctx, globalstate.EgoscaleV3Client, v3.ZoneName(c.Zone))
```
pierre-emmanuelJ, line 51: Feel free to change some type in all struct definition if you want to match egoscale v3 one, like for `ZoneName`, since everything will be v3

pierre-emmanuelJ: it could be applied to all the other type you see and think relevant to be updated

cmd/nlb_list.go
```diff
@@ -70,19 +71,22 @@ func (c *nlbListCmd) cmdRun(_ *cobra.Command, _ []string) error {
...
 		if err != nil {
 			return fmt.Errorf("unable to list Network Load Balancers in zone %s: %w", zone, err)
 		}
 
-		for _, nlb := range list {
+		for _, nlb := range list.LoadBalancers {
 			res <- nlbListItemOutput{
-				ID:        *nlb.ID,
-				Name:      *nlb.Name,
+				ID:        nlb.ID.String(),
```
pierre-emmanuelJ, line 86: like here for `ID`

tgrondier: While I conceptually agree and it's tempting to have the types of the output fields match the data struct fields, we end up typecasting those values as string anyway while building the table; i'll keep it for `Zone` for now as it's used when building the client

cmd/nlb_create.go
tgrondier: updated in 4abefb6f

arthuraliiev (approved):

kobajagi (approved):

PR #708, opened 2025-06-10
Update README.md

saiyam1814:
Current link points to 404


lucj (approved):
Nice catch. We changed the community site not long ago and left this 404 behind.

sauterp (approved):

PR #736, opened 2025-07-16
Remove last dependecies on egoscale v2

tgrondier:
# Description

This PR removes all the last dependencies of the exoscale client on egoscale v2

## Checklist
(For exoscale contributors)

* [x] Testing

## Testing

Test in individual commits

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -107,5 +107,3 @@ require (
 go 1.24
 
 toolchain go1.24.4
-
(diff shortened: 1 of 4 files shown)


cmd/internal/x/security_provider.go
```diff
@@ -1,4 +1,4 @@
-package api
+package x
```
tgrondier, line 1: Initially in `egoscale/v2`, this file is made obsolete by egoscale v3; however it's still useful for arbitrary calls such as the ones required by x, so it would make sense to move it here.

PR #750, opened 2025-10-04
feat: don't switch zone for dns command

mlec1:
# Description

The DNS command are not zone dependent, so this MR remove the need to change the zone for DNS command.

The reason behind that: I created a role for DNS challenge for getting web certificate. 
```
{
  "default-service-strategy": "deny",
  "services": {
    "dns": {
      "type": "rules",
      "rules": [
        {
          "expression": "parameters.has('type') && parameters.type != 'TXT'",
          "action": "deny"
        },
        {
          "expression": "parameters.has('name') && !parameters.name.startsWith('_acme-challenge')",
          "action": "deny"
        },
        {
          "expression": "resources.has('dns_domain_record') && resources.dns_domain_record.has('type') && resources.dns_domain_record.type != 'TXT'",
          "action": "deny"
        },
        {
          "expression": "resources.has('dns_domain_record') && resources.dns_domain_record.has('name') && !resources.dns_domain_record.name.startsWith('_acme-challenge')",
          "action": "deny"
        },
        {
          "expression": "operation in ['list-dns-domains', 'list-dns-domain-records', 'get-dns-domain-record', 'create-dns-domain-record', 'delete-dns-domain-record']",
          "action": "allow"
        }
      ]
    }
  }
}
```

I wanted to test if my role allows only the TXT records that I wanted. The easiest was to test with the exo cli.

However, due to the `SwitchClientZoneV3`, the role need also the permission `compute.list-zones` (or something similar), if I want to edit DNS Records with the exo cli, which is not needed for the tools like lego or cert-manager. 

From what I understood, the client will be created within the `buildClient`. Let me know if I missed something.

## Checklist
(For exoscale contributors)

* [ ] Testing

## Testing

Create, delete, update DNS records work.

### DIFF
--- a/cmd/dns/dns_list.go
+++ b/cmd/dns/dns_list.go
@@ -7,14 +7,12 @@ import (
 
 	"github.com/spf13/cobra"
 
-	// "github.com/exoscale/cli/pkg/account"
 	"github.com/exoscale/cli/utils"
 
 	exocmd "github.com/exoscale/cli/cmd"
 	"github.com/exoscale/cli/pkg/globalstate"
 	"github.com/exoscale/cli/pkg/output"
 	"github.com/exoscale/cli/table"
-	// v3 "github.com/exoscale/egoscale/v3"
 )
 
 type dnsListItemOutput struct {


cmd/dns/dns_list.go
```diff
@@ -7,14 +7,14 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/exoscale/cli/pkg/account"
+	// "github.com/exoscale/cli/pkg/account"
```
pierre-emmanuelJ: same here

PR #773, opened 2025-12-18
feat: add a warning in case the object doesn't exists before generating the presign url

mlec1:
# Description
It happened sometimes that when I want to generate a presigned url, I make a typo in the object name / path.

And it is only when I sent the link to someone or when I want to download the object, that I notice that the URL returns 404. 
I would prefer to notice it when I generate the URL. 

This MR adds a warning message when the object doesn't exists. For the moment, only a warning, I could return the error and no presigned url at all if you prefer. 

## Checklist
(For exoscale contributors)

* [x] Testing


imiric:
Thanks for the changes!

PR #808, opened 2026-03-04
feat(dbaas): add missing PostgreSQL configuration options

natalie-o-perret:
# Description

The CLI was missing 5 PostgreSQL configuration parameters that are already available in the API and SDK: `shared-buffers-percentage`, `synchronous-replication`, `timescaledb-settings`, `variant`, and `work-mem`. Users had no way to set these options through `exo dbaas create` or `exo dbaas update`, and `exo dbaas show` did not display them.

This PR adds full CLI support for all 5 parameters. `timescaledb-settings` is a JSON payload validated against the schema returned by `GetDBAASSettingsPG()`, following the same pattern as `pg-settings`, `pgbouncer-settings`, and `pglookout-settings`. 

The remaining 4 parameters (`shared-buffers-percentage`, `synchronous-replication`, `variant`, `work-mem`) are plain root-level flags passed directly to the request struct; no hardcoded client-side validation is added for them, consistent with how the API handles validation for values that may change upstream.

## Changes

* Added 5 new hidden flags to `dbaasServiceCreateCmd` and `dbaasServiceUpdateCmd`: `--pg-shared-buffers-percentage` (int64), `--pg-synchronous-replication` (string), `--pg-timescaledb-settings` (JSON), `--pg-variant` (string), `--pg-work-mem` (int64, MB).
* Wired the new flags into `createPG()` in `cmd/dbaas/dbaas_create_pg.go`. `timescaledb-settings` is validated against the schema returned by `GetDBAASSettingsPG()`, consistent with `pg-settings`, `pgbouncer-settings`, and `pglookout-settings`. The remaining 4 flags are plain root-level flags assigned directly to the request struct.
* Wired the new flags into `updatePG()` in `cmd/dbaas/dbaas_update_pg.go` using the `cmd.Flags().Changed()` pattern already used by every other update flag.
* Added 4 fields to the `exo dbaas show` output (`shared-buffers-percentage`, `synchronous-replication`, `timescaledb-settings`, `work-mem`). `variant` is not returned by the GET endpoint.
* Added `timescaledb` as a valid value for `--show-settings` in `exo dbaas show` and `--settings` in `exo dbaas type show`.

API reference: https://community.exoscale.com/reference/api/dbaas/dbaas-postgresql/#create-dbaas-service-pg

## Checklist
(For exoscale contributors)

* [x] Testing

## Testing

Build passes (`go build ./...`). Manual verification pending API access.

### DIFF
--- a/cmd/dbaas/dbaas_create_pg.go
+++ b/cmd/dbaas/dbaas_create_pg.go
@@ -119,13 +119,8 @@ func (c *dbaasServiceCreateCmd) createPG(_ *cobra.Command, _ []string) error {
 		databaseService.PGSettings = settings
 	}
 
-	if c.PGSharedBuffersPercentage != 0 {
-		databaseService.SharedBuffersPercentage = c.PGSharedBuffersPercentage
-	}
-
-	if c.PGSynchronousReplication != "" {
-		databaseService.SynchronousReplication = v3.EnumPGSynchronousReplication(c.PGSynchronousReplication)
-	}
+	databaseService.SharedBuffersPercentage = c.PGSharedBuffersPercentage
+	databaseService.SynchronousReplication = v3.EnumPGSynchronousReplication(c.PGSynchronousReplication)
 
 	if c.PGTimescaledbSettings != "" {
 		_, err := validateDatabaseServiceSettings(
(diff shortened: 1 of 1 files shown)


kobajagi:
`timescaledb-settings` is equivalent to `pg-settings` (and other `*-settings`) and it is inline json format with schema exposed by api (which we use for validation). All other parameters are root level and they are handled as normal flags so paragraph 2 of PR description is incorrect.

cmd/dbaas/dbaas_create_pg.go
```diff
@@ -119,6 +119,37 @@ func (c *dbaasServiceCreateCmd) createPG(_ *cobra.Command, _ []string) error {
 		databaseService.PGSettings = settings
 	}
 
+	if c.PGSharedBuffersPercentage != 0 {
```
kobajagi: Check is not needed, attribute is int64 and has `omitempty` tag.

PR #814, opened 2026-03-24
feat(storage): implement `move` command

### LINKED ISSUE #604: [Feature Request]: implement a `exo storage move`
### What feature would you like to have in the cli?

I accidentally uploaded a bunch of files to the wrong place in the bucket. As far as I can see there is currently no way to simply moving files to a different location within the same bucket. I could download them and then upload them again to the correct location but this is A) slow B) incurs egress costs/unnecessarily uses bandwidth. I would love for there to be some variation of `exo storage move <bucket>/<source> <bucket>/<dest>`. Being able to move between buckets would be a nice to have, but for me personally is not necessary. However, the point of this is mostly to be more efficient with egress/bandwidth, and I feel like there has to be a easier solution than download/reupload. I'm happy to assist in any way I can. 

Thanks!

nhedger:
> [!NOTE]
>  **Disclosure**: 

# Description

Closes #604

This PR adds `exo storage move` (`mv`) to move objects within a bucket or across buckets without downloading them locally.

The implementation uses **server-side copy + delete** for regular objects, and **multipart server-side copy for large objects**. It also **preserves object metadata, headers, and ACLs** across the move.

A few notes to help review:
- `cmd/storage/storage_move.go` adds the new CLI command, argument parsing, dry-run / verbose behavior, recursive prefix support, and a `--concurrency` flag for multipart copy workers
- `pkg/storage/sos/move.go` adds the storage-layer move implementation for both single-object and recursive moves, including multipart copy + abort-on-failure handling
- `pkg/storage/sos/acl.go` refactors ACL translation so the same ACL preservation logic can be reused for both `CopyObject` and `CreateMultipartUpload`
- `pkg/storage/sos/object.go` extracts multipart part-size calculation into a shared helper so upload and move use the same dynamic sizing logic and stay under the 10,000-part limit
- `pkg/storage/sos/s3api.go`, `pkg/storage/sos/s3api_mock_test.go`, and `pkg/storage/sos/move_test.go` extend the S3 abstraction and test coverage for `HeadObject`, multipart copy, abort handling, and configurable concurrency

## Open questions

- [ ] Should `--concurrency` remain a user-facing flag for multipart moves? While it can improve large-object move times, higher values may put additional strain on the object storage backend and increase request pressure on the infrastructure.

## Checklist
(For exoscale contributors)

* [x] Testing

## Testing
- `go test ./...`
- verified dry-run behavior for single-object moves
- verified same-bucket object moves
- verified cross-bucket object moves
- verified recursive cross-bucket prefix moves
- verified live `>5 GiB` cross-bucket move path
- verified configurable multipart copy concurrency

### DIFF
--- a/cmd/storage/storage_move.go
+++ b/cmd/storage/storage_move.go
@@ -1,0 +1,188 @@
+package storage
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	exocmd "github.com/exoscale/cli/cmd"
+	"github.com/exoscale/cli/pkg/globalstate"
+	"github.com/exoscale/cli/pkg/storage/sos"
+)
+
+type storageMoveLocation struct {
+	Bucket string
+	Key    string
+}
+
+var storageMoveCmd = &cobra.Command{
+	Use:     "move SOURCE DESTINATION",
+	Aliases: []string{"mv"},
+	Short:   "Move objects without local download",
+	Long: `This command moves objects between buckets without downloading them locally.
+
+Examples:
+
+    # Move an object within a bucket
+    exo storage move sos://my-bucket/file-a sos://my-bucket/archive/file-a
+
+    # Move an object to another bucket and keep its basename
+    exo storage move sos://my-bucket/file-a sos://other-bucket/archive/
+
+    # Move a prefix recursively
+    exo storage move -r sos://my-bucket/public/ sos://other-bucket/archive/
+
+Notes:
+
+    * If the destination ends with "/", the source basename is preserved.
+    * Prefix moves preserve relative paths under the destination prefix.
+    * Existing destination objects are overwritten.
+`,
+
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) != 2 {
+			exocmd.CmdExitOnUsageError(cmd, "invalid arguments")
+		}
+
+		args[0] = strings.TrimPrefix(args[0], sos.BucketPrefix)
+		args[1] = strings.TrimPrefix(args[1], sos.BucketPrefix)
+
+		if !strings.Contains(args[0], "/") {
+			exocmd.CmdExitOnUsageError(cmd, fmt.Sprintf("invalid argument: %q", args[0]))
+		}
+
+		return nil
+	},
+
+	RunE: func(cmd *cobra.Command, args []string) error {
+		source, err := parseStorageMoveLocation(args[0])
+		if err != nil {
+			return err
+		}
+
+		destination, err := parseStorageMoveLocation(args[1])
+		if err != nil {
+			return err
+		}
+
+		recursive, err := cmd.Flags().GetBool("recursive")
+		if err != nil {
+			return err
+		}
+
+		dryRun, err := cmd.Flags().GetBool("dry-run")
+		if err != nil {
+			return err
+		}
+
+		verbose, err := cmd.Flags().GetBool("verbose")
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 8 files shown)


natalie-o-perret:
Thanks for this PR!

The implementation is solid overall, clean separation of concerns, and proper abort handling for multipart uploads.

A couple of things that might be worth addressing:

1. **Documenting the non-atomic copy+delete behaviour**
   - `CopyObject` + `DeleteObjects` is not atomic. If the delete fails after a successful copy, the object will exist in both the source and destination with no automatic rollback.
   - This is a "generic" S3 limitation, but it might be worth warning users about it.
   - Could be a small note in the `Long` description in [`cmd/storage/storage_move.go` L23](https://github.com/exoscale/cli/blob/d9e5985f949e573bcb2ad188ae72ec2fc2ecd29d/cmd/storage/storage_move.go#L23), e.g.,
     > Warning: move is implemented as server-side copy followed by delete.
     >
     > If the delete step fails, the object will remain in both locations.
     >
     > There is no automatic rollback.

2. **`--concurrency` vs `--multipart-concurrency`**
   - The flag controls the number of workers for multipart copy parts within a single large object, not how many objects are moved in parallel, the outer loop in [`MoveObjects` L62](https://github.com/exoscale/cli/blob/d9e5985f949e573bcb2ad188ae72ec2fc2ecd29d/pkg/storage/sos/move.go#L62) is serial.
   - `--multipart-concurrency` might reflect that more accurately and avoid confusion for users with large buckets of small objects.
   - Could also be worth mentioning in the `Long` description that multi-object prefix moves are processed serially.

3. **`-r` vs trailing `/`**
   - A trailing `/` on the source already triggers prefix mode, `-r` only controls whether `ForEachObject` recurses into subdirectories.
   - `exo storage move sos://bucket/prefix/ sos://other/` (without `-r`) would do a non-recursive prefix move, which might not be obvious.
   - A small clarification in the `Long` description or flag usage could help.

4. **Tests for the multipart copy path**
   - [`move_test.go` L19](https://github.com/exoscale/cli/blob/d9e5985f949e573bcb2ad188ae72ec2fc2ecd29d/pkg/storage/sos/move_test.go#L19) covers the single-object path nicely.
   - The multipart path in [`copyMultipartObjectToDestination` L252](https://github.com/exoscale/cli/blob/d9e5985f949e573bcb2ad188ae72ec2fc2ecd29d/pkg/storage/sos/move.go#L252), the worker pool, part assembly, cancel-on-error, abort-on-failure, might be worth covering too, even just a happy-path and a failure/abort case.

5. **Minor: comment in `deleteMovedObject`**
   - [`deleteMovedObject` L398](https://github.com/exoscale/cli/blob/d9e5985f949e573bcb2ad188ae72ec2fc2ecd29d/pkg/storage/sos/move.go#L398) only inspects `res.Errors[0]`, which is fine since it always deletes exactly one object, a short `// always one object` comment might just make that obvious at a glance.

cmd/storage/storage_move.go
```diff
@@ -0,0 +1,188 @@
...
+				fmt.Printf("no objects exist at %q\n", source.Key)
+			}
+			return nil
+		}
+
+		if dryRun {
+			return nil
+		}
+
+		return nil
```
kobajagi, line 153:
```suggestion
		if len(moved) == 0 && !globalstate.Quiet {
			fmt.Printf("no objects exist at %q\n", source.Key)
		}
		
		return nil
```

natalie-o-perret:
Thank you so much for this implementation and the detailed review feedback! 

We really appreciate the effort that went into it. 

We're going to close this PR and implement the move feature in a separate PR to ensure it stays consistent with our existing codebase patterns. 

Excited to have this feature land soon!

PR #816, opened 2026-03-25
feat(compute): add `--reverse-dns` flag to create command

### LINKED ISSUE #564: [Feature Request]: add `--reverse-dns string` to `compute instance create`
### What feature would you like to have in the cli?

The `compute instance create` command is missing an argument to set the reverse DNS domain. Such an argument is already available in `compute instance update`:

```
> exo compute instance update --help
This command updates an Instance .
[...]
Flags:
[...]
      --reverse-dns string     Reverse DNS Domain
```

nhedger:
> [!NOTE]
>  **Disclosure**: 

# Description

Closes #564

This PR adds a `--reverse-dns` flag to the `compute instance create` command.

The implementation follows the same approach already used by `compute instance update`: after the instance is created and the operation completes successfully, the CLI applies reverse DNS through the dedicated reverse-DNS API endpoint before returning the final instance output.

Test coverage was added for both the CLI surface and the API-backed lifecycle:
- a without-API testscript scenario checks that `compute instance create --help` exposes `--reverse-dns`
- a with-API testscript scenario creates an instance with reverse DNS, verifies it via `compute instance show`, then deletes it
One API detail worth noting is that the returned reverse DNS value includes a trailing dot, so the API test matches that behavior.

## Checklist
(For exoscale contributors)

* [x] Testing

## Testing

- `go build -mod=vendor -o bin/exo .`
- `go test ./cmd/...`
- `go test -C tests/e2e -run 'TestScriptsLocal/compute_instance_create_flags' -v`
- `go test -C tests/e2e -tags=api -run 'TestScriptsAPI/instance_create_reverse_dns' -v`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - sks: add `rotate-karpenter-credentials` command #797
 - sks: add `active-nodepool-templates` command #797
 - new command `exo ai deployment instance-type` that allows showing what GPU is usable in which zone #809
+- instance create: add support for setting a reverse dns with the `--reverse-dns` flag
 
 ### Bug fixes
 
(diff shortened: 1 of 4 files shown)


imiric (approved):
Thanks for this! 🙇 

The implementation LGTM, and the tests are appreciated 👍 They pass locally for me.

natalie-o-perret (approved):

PR #817, opened 2026-03-25
fix(config): improve config override behavior

### LINKED ISSUE #737: [Bug]: Default zone is not respected
### Current Behavior

When listing e.g. block storage resources, this is done for all zones:

`exo c bs list
`

gives correctly a list of all block storage resources throughout all zones.

If I copy one of the IDs of a block storage of the default zone, which is de-fra-1 in my case, i get:

> error: "pvc-XYZ" not found in ListBlockStorageVolumesResponse: Not Found

If I add --zone de-fra-1, the volume is shown correctly.

In my exoscale.toml there is a line: `defaultZone = 'de-fra-1'` The credentials work correctly, so it's the correct file.

### Expected Behavior

The details of the volume should be shown, if the volume is located in the default zone configured in my exoscale.toml

### Steps To Reproduce

1. Create a block storage in de-fra-1
2. Configure the default zone for the CLI to de-fra-1
3. List all block storages
4. try to get the details of the one created in the first step without specifying a --zone parameter

### CLI Version

exo 1.85.3 3130c6ae (egoscale 0.102.3)

### Os platform

MacOS

### Relevant log output

```shell

```

nhedger:
> [!NOTE]
>  **Disclosure**: 

# Description

Fixes #737 

This updates config resolution so environment variables override account credentials without discarding the rest of the selected profile.

Previously, when `EXOSCALE_API_KEY` and `EXOSCALE_API_SECRET` were set, the CLI returned early during config initialization and skipped loading the config file entirely. As a result, profile values such as `defaultZone` were ignored and commands relying on the account default zone could fall back to the built-in default instead of the zone configured in `exoscale.toml`.

This change now:
- loads and selects the config profile first when a config file is available
- applies environment overrides afterwards for credentials and related override fields
- preserves profile settings such as `defaultZone` and other account defaults
- keeps the existing env-only behavior when no config file is present
This fixes the override precedence so env credentials replace authentication settings, not the whole profile.

## Checklist
(For exoscale contributors)

* [x] Testing

## Testing

- Ran `go test ./...`
- Ran targeted e2e coverage for the config/env precedence:
  - `go test -run 'TestScriptsLocal/(show_with_env_credentials_keeps_config_defaults|show_with_env_credentials_without_config)$' -count=1 ./...` from `tests/e2e`
- Verified that:
  - env credentials still work without a config file
  - `defaultZone` from the config is preserved when env credentials are set
- Confirmed that #737 cannot be reproduced anymore:
  ```sh
  export EXOSCALE_API_KEY="..."
  export EXOSCALE_API_SECRET="..."
  VOL_ID=$(./bin/exo --output-format json compute block-storage create repro-737-$(date +%s) --zone ch-gva-2 --size 10 | jq -r '.id')
  ./bin/exo compute block-storage list
  ./bin/exo compute block-storage show "$VOL_ID"
  ./bin/exo compute block-storage show "$VOL_ID" --zone ch-gva-2
  ./bin/exo compute block-storage delete "$VOL_ID" --zone ch-gva-2 -f
  ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - fix(nlb): API error swallowed on load-balancer update (e.g. duplicate name conflict reported as "operation is nil") #806
 - fix(config): panic when used without a default account set #798
 - Fix bad flag ref in `dns add NS` #812
+- fix(config): preserve profile defaults such as default zone when API credentials are provided via environment variables 
 
 ### Documentation
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 - fix(nlb): API error swallowed on load-balancer update (e.g. duplicate name conflict reported as "operation is nil") #806
 - fix(config): panic when used without a default account set #798
 - Fix bad flag ref in `dns add NS` #812
-- fix(config): preserve profile defaults such as default zone when API credentials are provided via environment variables 
+- fix(config): preserve profile defaults such as default zone when API credentials are provided via environment variables #817
 
 ### Documentation
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -107,6 +107,120 @@ func init() {
 
 var ignoreClientBuild = false
 
+type envAccountOverrides struct {
+	apiEndpoint    string
+	apiEnvironment string
+	apiKey         string
+	apiSecret      string
+	sosEndpoint    string
+	clientTimeout  *int
+}
+
+func readEnvAccountOverrides() envAccountOverrides {
+	clientTimeoutFromEnv := readFromEnv("EXOSCALE_API_TIMEOUT")
+
+	overrides := envAccountOverrides{
+		apiEndpoint:    os.Getenv("EXOSCALE_API_ENDPOINT"),
+		apiEnvironment: readFromEnv("EXOSCALE_API_ENVIRONMENT"),
+		apiKey: readFromEnv(
+			"EXOSCALE_API_KEY",
+			"EXOSCALE_KEY",
+			"CLOUDSTACK_KEY",
+			"CLOUDSTACK_API_KEY",
+		),
+		apiSecret: readFromEnv(
+			"EXOSCALE_API_SECRET",
+			"EXOSCALE_SECRET",
+			"EXOSCALE_SECRET_KEY",
+			"CLOUDSTACK_SECRET",
+			"CLOUDSTACK_SECRET_KEY",
+		),
+		sosEndpoint: readFromEnv(
+			"EXOSCALE_STORAGE_API_ENDPOINT",
+			"EXOSCALE_SOS_ENDPOINT",
+		),
+	}
+
+	if clientTimeoutFromEnv != "" {
+		if t, err := strconv.Atoi(clientTimeoutFromEnv); err == nil {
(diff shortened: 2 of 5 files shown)


imiric (commented):
Hi, thanks for the stream of improvements and fixes you've contributed! It's much appreciated :bow: I also appreciate the additional tests, and disclosure of "AI" use. :thumbsup:

I confirmed that this PR fixes the issue described in #737, which is great. However, I'm not sure if this is the right approach. :thinking:

I don't think implementing manual overrides like this is a robust or maintainable solution. The canonical way to ensure that values from CLI flags, env vars, config files, and other sources, have a correct and predictable order of precedence within a Cobra application is to use [the related Viper library, which has a commonly accepted precedence order](https://github.com/spf13/viper?tab=readme-ov-file#putting-values-in-viper).

We currently only use it for reading the config file, and not much else. So the previous code wasn't great either, but I think we should leverage Viper more instead of doing this overriding ourselves. This way we can align with a common convention, it can work for other values besides the ones we specified here (e.g. [there is interest in supporting `EXOSCALE_ZONE`](https://github.com/exoscale/cli/issues/737#issuecomment-3772897841)), and it should likely be less code for us to maintain, and therefore hopefully less chances for surprising behavior and bugs.

For example, even on this branch, the `EXOSCALE_CONFIG` env var takes precedence over the `--config` flag, which shouldn't be the case.

I'm not sure how difficult it would be to partially improve our usage of Viper to fix this particular issue without having to refactor the rest of the code base. Ideally that could get us on the right path to expand the improvement for other commands and values later.

So I don't want to block this from merging, but would strongly suggest revisiting this fix with the above in mind. I'll let others from @exoscale/giza chime in.

nhedger:
Thanks for your review. I wasn't sure if you'd want such a refactor to begin with, so I went for the quick win, but I agree it would probably be more robust to use Viper.

I'm happy to explore an alternative way using Viper. Just let me know if you you still want to merge this in the meantime, in which case I'd create another PR for te Viper refactor

Cheers

natalie-o-perret:
Hey there,

@imiric, you're right that the manual override pattern isn't great long-term, and the `EXOSCALE_CONFIG vs --config` catch is a good one.

@nhedger, before going down the Viper path for the follow-up, worth noting that (based on my understanding of) Viper's precedence model maps flat key-value pairs. It doesn't natively handle our "select a profile from a list, then overlay env vars on top of that profile's fields" pattern. Imho, we might still need custom merge logic for the multi-account model; the question is just whether Viper or plain `os.Getenv` feeds into it.

What I think could be a wee more sustainable (in a dedicated follow-up PR) is separating source loading from resolution: read env vars, config file, and flags into discrete structs independently, then merge them with a single explicit-precedence function. That would make the rules visible as code and easy to unit-test. Adding something like `EXOSCALE_ZONE` later becomes one new field instead of scattered `if`-chains.

The `envAccountOverrides` struct in this PR is actually a decent first step toward that pattern. 

---

Good to merge for the #737 fix, it still fixes a real bug. 

That said, the config resolution logic could really use a rethink / revamp. 

Happy to help shape that.

natalie-o-perret (approved):

pierre-emmanuelJ (approved):

natalie-o-perret:
Merging, and I will revamp the conf logic in a subsequent PR.

PR #821, opened 2026-04-02
fix: print API error message instead of operation nil

mlec1:
# Description
Exactly like in https://github.com/exoscale/cli/pull/806, it returns `operation is nil` when update a SKSNodePool when it is called with parameters that the API rejects.

`exo compute sks nodepool update itops-test itops-test-workers --label managed_uptime=standard --label dont_manage_until=2026-05-06 --label test=test/test`

image

At the opposite of https://github.com/exoscale/cli/pull/806, I didn't add a e2e test because I encountered the issue while trying to add a `/` in a label of a SKS Nodepool but at the end I would like that it works (I made a support ticket for this), so if I make a E2E test for it now, it might not pass anymore once `/` are accepted on the API Side.

## Checklist
(For exoscale contributors)

* [ ] Changelog updated (under *Unreleased* block, and add the Pull Request #number for each bit you add to the `CHANGELOG.md`)
* [ ] Testing

## Testing

image

Now returns the BAD Request response from the API.

### DIFF
--- a/cmd/compute/sks/sks_nodepool_update.go
+++ b/cmd/compute/sks/sks_nodepool_update.go
@@ -203,6 +203,9 @@ func (c *sksNodepoolUpdateCmd) CmdRun(cmd *cobra.Command, _ []string) error { //
 
 	if updated {
 		op, err := client.UpdateSKSNodepool(ctx, cluster.ID, nodepool.ID, updateReq)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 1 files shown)


kobajagi (approved):

pierre-emmanuelJ (approved):

imiric (approved):
Thanks!

PR #825, opened 2026-04-24
improve error management for RFC-9457 compliant errors

habfast:
# Description

Follow-up on this: https://github.com/exoscale/egoscale/pull/763

Shortcut ticket: [sc-176689](https://app.shortcut.com/exoscale/story/176689/improve-error-management-of-cli)

## Checklist
(For exoscale contributors)

* [x] Testing

## Testing

Screenshot 2026-04-24 at 16 44 10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - fix(nlb): API error swallowed on load-balancer update (e.g. duplicate name conflict reported as "operation is nil") #806
 - fix(config): panic when used without a default account set #798
 - Fix bad flag ref in `dns add NS` #812
+- Improved error reporting for Dedicated Inference #825
 
 ### Documentation
 
--- a/vendor/github.com/exoscale/egoscale/v3/errors.go
+++ b/vendor/github.com/exoscale/egoscale/v3/errors.go
@@ -97,8 +97,15 @@ var httpStatusCodeErrors = map[int]error{
 func handleHTTPErrorResp(resp *http.Response) error {
 	if resp.StatusCode >= 400 && resp.StatusCode <= 599 {
 		var res struct {
+			// Simple error formats
 			Message string `json:"message"`
 			Error   string `json:"error"`
+			// RFC 7807 Problem Details format
+			Title  string `json:"title"`
+			Errors []struct {
+				Detail   string `json:"detail"`
+				Location string `json:"location"`
+			} `json:"errors"`
 		}
 
 		data, err := io.ReadAll(resp.Body)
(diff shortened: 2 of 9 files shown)


vendor/github.com/exoscale/egoscale/v3/schemas.go
```diff
@@ -74,6 +74,26 @@ type AccessKeyResource struct {
...
+	// Organization UUID that owns this key
+	OrgUuid UUID `json:"org-uuid,omitempty"`
+	// Key scope: 'public' for all deployments, or a specific deployment UUID
+	Scope string `json:"scope,omitempty"`
+	// Last update timestamp
+	UpdatedAT time.Time `json:"updated-at,omitempty"`
+}
+
+// AI API key with plaintext value
+type AIAPIKeyWithValue struct {
```
natalie-o-perret, line 94: This `struct` is empty, could it be a generator issue?

The key value can't be never captured 🤔

CHANGELOG.md
```diff
@@ -33,6 +17,7 @@
 - fix(nlb): API error swallowed on load-balancer update (e.g. duplicate name conflict reported as "operation is nil") #806
 - fix(config): panic when used without a default account set #798
 - Fix bad flag ref in `dns add NS` #812
+- Improved error reporting for Dedicated Inference #825
```
natalie-o-perret, line 20: It doesn't seem to be only about Dedicated Inference, wdyt?

vendor/github.com/exoscale/egoscale/v3/errors.go
```diff
@@ -94,11 +94,41 @@ var httpStatusCodeErrors = map[int]error{
...
+}
+
+// APIErrorEntry is one item from the RFC 9457 "errors" array.
+// Location is a JSONPath expression (e.g. "$['some-field']") when the server
+// provides field-level detail; it is empty for plain string error entries.
+type APIErrorEntry struct {
+	Detail   string
+	Location string
+}
+
```
natalie-o-perret, line 120: Nit-ish, 

```go
func (e *APIError) Error() string { return e.sentinel.Error() }
```

Only returns the sentinel, callers that just do `err.Error()` (e.g. log lines, other middleware in the CLI) could get only "not found" with no detail. 

Detail is only surfaced if the error flows through `formatError`. 

Worth at least noting in a comment so future contributors don't wonder why detail is "missing" in logs.

```suggestion
// Error returns only the HTTP status sentinel. Use errors.As(*APIError) to access structured detail.
```

natalie-o-perret (commented):
~~Nice improvement, but it might be worth addressing the comments I've left.~~

[Edit]
Approved, (we should address the Egoscale generation at some point).

CHANGELOG.md
habfast: At this time it is, because Dedicated Inference is the only product in production that implements RFC-9457 compliant errors. KMS is coming soon too, but it hasn't been released yet. More will come, but in the current state of affairs, this patch only improves things for my service.

vendor/github.com/exoscale/egoscale/v3/schemas.go
habfast: good point... @loicrouillermonay 

in any case, this is library code, not directly related to this PR...

vendor/github.com/exoscale/egoscale/v3/errors.go
habfast: this is egoscale code - I can go back to the lib and fix it, but it's outside the scope of this PR to change egoscale...

natalie-o-perret (approved):

sauterp (approved):

PR #826, opened 2026-04-27
Improve zone failure handling

habfast:
# Description

Last Friday the network in ZAG1 went down and `exo ai deployment list` (and every other multi-zone list) stopped responding for everyone. This bricked the whole command, because the fanout waited on every zone and a single error nuked the buffered output before render.
This PR is a proposal to fix that, scoped to `exo ai deployment list` as the proving ground. The plan is to migrate the rest of the listers in follow-ups now that the helpers are in place.

Shortcut story: [sc-177228](https://app.shortcut.com/exoscale/story/177228/improve-zone-failure-handling)

## What changed

Screenshot 2026-04-27 at 17 16 41

<br>
Screenshot 2026-04-27 at 17 17 00
<br>
Screenshot 2026-04-27 at 17 23 32
<br>

- New `utils.ForEveryZoneAsync` replacement for `ForEveryZone`, with a per-zone timeout and a buffered warning sink instead of multierror. One bad zone produces a stderr warning and returns. Comes with a `--timeout` global flag (default 15s, env `EXOSCALE_TIMEOUT`).
- Streaming output. Rows render progressively as each zone returns rather than after the slowest one finishes. Table + text stream row-by-row; JSON stays buffered (one valid array). New `output.StreamingOutputter` interface in `pkg/output/streaming.go`.
- Bordered streaming table. Matches the existing tablewriter look (same `┼` `│` `─` glyphs, same one-space cell padding, uppercase headers). Column widths are pre-specified per row-struct field via a new `outputWidth:"36"` struct tag. We can't auto-size when we don't know how many rows are coming.
- Warnings are buffered. Per-zone errors collect in a `WarningSink` and flush at the end (or on Ctrl+C), so streamed rows don't get interleaved with stderr noise.
- Liveness glyph. A single rotating braille char on stderr while the `utils.ForEveryZoneAsync` is running, so the user knows things aren't hung even when fast zones have already streamed and we're still waiting on a slow one. Silent on non-TTY and under `--quiet`. 
- Exit code 1 on partial failure, so scripts can tell.

## Testing

- Unit tests for the new helpers
- Manually verified against preprod (3 of 4 zones currently down). With this branch you get the healthy zone's rows immediately and three `warning: zone X: ...` lines on stderr at the end.

## Reviewer notes

- Review commit by commit - the history is pretty clean
- The streamed table will look slightly different from the `tablewriter`-rendered one if a value is wider than its declared `outputWidth` the cell trims the content. I hesitated between that and letting the content expand, but this seemed the better tradeoff, it's more readable.
- The single-glyph spinner can theoretically interleave bytes with a row write if stderr/stdout are racing, but the glyph is one char and ticks every 100ms. In practice the row's leading `│` cleanly covers it.

## Checklist
(For exoscale contributors)

* [x] Testing

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -15,6 +15,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -158,6 +159,7 @@ func init() {
 	RootCmd.PersistentFlags().StringVarP(&globalstate.OutputFormat, "output-format", "O", "", "Output format (table|json|text), see \"exo output --help\" for more information")
 	RootCmd.PersistentFlags().StringVar(&output.GOutputTemplate, "output-template", "", "Template to use if output format is \"text\"")
 	RootCmd.PersistentFlags().BoolVarP(&globalstate.Quiet, "quiet", "Q", false, "Quiet mode (disable non-essential command output)")
+	RootCmd.PersistentFlags().DurationVar(&globalstate.RequestTimeout, "timeout", 15*time.Second, "Per-zone timeout for list operations [env EXOSCALE_TIMEOUT]")
 	RootCmd.AddCommand(versionCmd)
 
 	// Don't attempt to load client configuration in testing mode.
@@ -175,6 +177,7 @@ func initConfig() { //nolint:gocyclo
 	envs := map[string]string{
 		"EXOSCALE_CONFIG":  "config",
 		"EXOSCALE_ACCOUNT": "use-account",
+		"EXOSCALE_TIMEOUT": "timeout",
 	}
 
 	for env, flag := range envs {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -46,6 +46,12 @@ var RootCmd = &cobra.Command{
 	Short:         "Manage your Exoscale infrastructure easily",
 	SilenceUsage:  true,
 	SilenceErrors: true,
+	PersistentPreRunE: func(_ *cobra.Command, _ []string) error {
+		if globalstate.RequestTimeout != -time.Second && globalstate.RequestTimeout <= 0 {
+			return fmt.Errorf("--timeout must be a positive duration (e.g. 15s), or -1s to disable")
+		}
+		return nil
+	},
 }
 
 var versionCmd = &cobra.Command{
@@ -159,7 +165,7 @@ func init() {
 	RootCmd.PersistentFlags().StringVarP(&globalstate.OutputFormat, "output-format", "O", "", "Output format (table|json|text), see \"exo output --help\" for more information")
 	RootCmd.PersistentFlags().StringVar(&output.GOutputTemplate, "output-template", "", "Template to use if output format is \"text\"")
 	RootCmd.PersistentFlags().BoolVarP(&globalstate.Quiet, "quiet", "Q", false, "Quiet mode (disable non-essential command output)")
-	RootCmd.PersistentFlags().DurationVar(&globalstate.RequestTimeout, "timeout", 15*time.Second, "Per-zone timeout for list operations [env EXOSCALE_TIMEOUT]")
+	RootCmd.PersistentFlags().DurationVar(&globalstate.RequestTimeout, "timeout", 15*time.Second, "Per-zone timeout for list operations; -1s disables timeout [env EXOSCALE_TIMEOUT]")
 	RootCmd.AddCommand(versionCmd)
 
 	// Don't attempt to load client configuration in testing mode.
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -11,6 +11,9 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/hashicorp/go-multierror"
 
@@ -234,6 +237,35 @@ func ForEveryZone(zones []v3.Zone, f func(zone v3.Zone) error) error {
 	}
 
 	return meg.Wait().ErrorOrNil()
+}
+
+// ForEveryZoneAsync runs f concurrently per zone with a per-zone timeout
+// derived from ctx. Per-zone errors are buffered into sink as warnings
+// and never abort the other zones. Returns the number of zones that
+// failed.
+func ForEveryZoneAsync(
+	ctx context.Context,
+	zones []v3.Zone,
+	timeout time.Duration,
+	sink *WarningSink,
+	f func(ctx context.Context, zone v3.Zone) error,
+) int {
+	var wg sync.WaitGroup
+	var failed atomic.Int32
+	for _, zone := range zones {
+		wg.Add(1)
+		go func(zone v3.Zone) {
+			defer wg.Done()
+			zCtx, cancel := context.WithTimeout(ctx, timeout)
+			defer cancel()
+			if err := f(zCtx, zone); err != nil {
+				failed.Add(1)
+				sink.Add("zone %s: %v", zone.Name, err)
+			}
+		}(zone)
+	}
+	wg.Wait()
+	return int(failed.Load())
 }
 
 // ParseInstanceType returns an v3.InstanceType with family and name.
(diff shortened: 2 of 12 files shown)


cmd/root.go
```diff
@@ -158,6 +159,7 @@ func init() {
 	RootCmd.PersistentFlags().StringVarP(&globalstate.OutputFormat, "output-format", "O", "", "Output format (table|json|text), see \"exo output --help\" for more information")
 	RootCmd.PersistentFlags().StringVar(&output.GOutputTemplate, "output-template", "", "Template to use if output format is \"text\"")
 	RootCmd.PersistentFlags().BoolVarP(&globalstate.Quiet, "quiet", "Q", false, "Quiet mode (disable non-essential command output)")
+	RootCmd.PersistentFlags().DurationVar(&globalstate.RequestTimeout, "timeout", 15*time.Second, "Per-zone timeout for list operations [env EXOSCALE_TIMEOUT]")
```
natalie-o-perret: Banana question of mine, but what if someone is passing `0`?
Should we have a guard for that? or treat `0` a no-timeout?

utils/utils.go
```diff
@@ -236,6 +239,47 @@ func ForEveryZone(zones []v3.Zone, f func(zone v3.Zone) error) error {
...
+		defer spinner.Stop()
+	}
+
+	var wg sync.WaitGroup
+	var failed atomic.Int32
+	for _, zone := range zones {
+		wg.Add(1)
+		go func(zone v3.Zone) {
+			defer wg.Done()
+			zCtx, cancel := context.WithTimeout(ctx, timeout)
```
natalie-o-perret: Echoing https://github.com/exoscale/cli/pull/826/changes#r3168564041

utils/warnings.go
```diff
@@ -0,0 +1,83 @@
...
+	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
+	done := make(chan struct{})
+	go func() {
+		select {
+		case sig := <-ch:
+			s.Flush()
+			signal.Stop(ch)
+			// Re-raise so the default handler terminates the process
+			// with the conventional exit code.
+			_ = syscall.Kill(syscall.Getpid(), sig.(syscall.Signal))
```
natalie-o-perret: I'm not sure that `syscall.Kill` is gonna work out for Windows OS's

- https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/syscall/zsyscall_darwin_amd64.go;l=402
- https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/syscall/zsyscall_linux_amd64.go;l=610
- https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/syscall/zsyscall_windows.go

While Windows is skipped for tests [here](https://github.com/exoscale/cli/pull/826/changes#diff-42d685424866f20e83de4d713cb6a4cefebdc92f8cf168e25a09915f0100fd39R153)
```go
if runtime.GOOS == "windows" {
	t.Skip("SIGINT semantics differ on Windows")
}
```

What about the production code? 

That would mean something like below?
```go
//go:build !windows`
```

Which we can't afford.

pierre-emmanuelJ: CLI is supported on Windows and still need to be supported on Windows

natalie-o-perret: Sorry, was supposed to be rhetorical, rephrased.

My point was that it seems that we need an alternative for that execution path in non-test / production code so that the behaviour is consistent across operating systems.

If `syscall.Kill` is deemed really necessary for unix os, we could down that kind of road:
```
- utils/warnings_unix.go (//go:build !windows) 
- utils/warnings_windows.go (//go:build windows) // still need to find an alternative, Interrupt?
```

But, tbf, I think we could use a cross-platform alternative like below:
```go
case sig := <-ch:
    s.Flush()
    signal.Stop(ch)
    if sc, ok := sig.(syscall.Signal); ok {
        os.Exit(128 + int(sc))
    }
    os.Exit(1)
```

---

Btw, just showing I'm not coming up with a claim that would be plucked out of thin air, here is the evidence that got me thinking:

```bash
$ GOOS=windows GOARCH=amd64 go build ./...
# github.com/exoscale/cli/utils
utils/warnings.go:74:16: undefined: syscall.Kill
exit: 1

$ GOOS=linux GOARCH=amd64 go build ./...
exit: 0

$ GOOS=darwin GOARCH=arm64 go build ./...
exit: 0
```

Side note: we should add a CI step that cross-compiles for all release targets, so issues like this can get caught automatically on PRs, rather than at release time.

cmd/root.go
habfast: I added some handling of that. Also added special value `-1s` to disable timeout.

utils/warnings.go
habfast: Thanks for the catch, I haven't very often coded for windows 😝

- This was a bit of a bonus, not a core feature
- I have no way of testing on windows

Given this, I just dropped the code and feature altogether. It's fine as it is. 

It now builds:
```
(.venv) habfast@mbp.local:~/workspace/cli (improve-zone-failure-handling*) $ GOOS=windows GOARCH=amd64 go build ./...

(.venv) habfast@mbp.local:~/workspace/cli (improve-zone-failure-handling*) $ echo $?
0
```

utils/utils.go
habfast: I think this is obsolete? Feel free to reopen if I missed something

natalie-o-perret: Yes, resolved by the fix in `cmd/root.go`, thanks :bowing_woman:.

cmd/root.go
natalie-o-perret: Thanks, the `-1s` convention is a nice convention for timeouts. :bowing_woman: 

---

One thing to be aware of, tho,: `PersistentPreRunE` on `RootCmd` won't run for subcommands that define their own, like `storageCmd` and `xCmd`. 

The `timeout > 0` guard inside `ForEveryZoneAsync` is still the actual safety net so nothing really breaks, just worth keeping in mind when migrating the other list commands :eyes:

utils/warnings.go
natalie-o-perret: Makes sense to drop it for now.

Just flagging that it's a quiet behavior regression: if the user Ctrl+Cs mid-fanout, `defer sink.Flush()` in `deployment_list.go` won't execute since Go's default SIGINT handler terminates before defers run, so buffered warnings would be lost. 

(I think my suggestion wasn't too bad, but I'm done "bikeshedding").

```go
package main

import (
	"fmt"
	"os"
	"os/signal"
	"syscall"
	"time"
)

var warnings = []string{
	"zone ch-gva-2: deadline exceeded",
	"zone de-fra-1: connection refused",
}

func flush() {
	for _, w := range warnings {
		fmt.Fprintf(os.Stderr, "warning: %s\n", w)
	}
}

// runOld: SIGINT caught, warnings flushed, then re-raised.
func runOld() {
	fmt.Fprintln(os.Stderr, "[old] fanout running... (send SIGINT now)")
	ch := make(chan os.Signal, 1)
	signal.Notify(ch, syscall.SIGINT)
	go func() {
		sig := <-ch
		flush()
		signal.Stop(ch)
		_ = syscall.Kill(syscall.Getpid(), sig.(syscall.Signal))
	}()
	defer flush()
	time.Sleep(10 * time.Second)
}

// runCurrent: no signal handler, defer never runs on SIGINT.
func runCurrent() {
	fmt.Fprintln(os.Stderr, "[current] fanout running... (send SIGINT now)")
	defer flush()
	time.Sleep(10 * time.Second)
}

func main() {
	if len(os.Args) < 2 {
		fmt.Fprintln(os.Stderr, "usage: demo [old|current]")
		os.Exit(1)
	}
	switch os.Args[1] {
	case "old":
		runOld()
	case "current":
		runCurrent()
	default:
		fmt.Fprintln(os.Stderr, "usage: demo [old|current]")
		os.Exit(1)
	}
}
```

```
# current: Ctrl+C after 0.5s
fanout running...

# old: Ctrl+C after 0.5s
fanout running...
warning: zone ch-gva-2: deadline exceeded
warning: zone de-fra-1: connection refused
```

natalie-o-perret (approved):

pierre-emmanuelJ (approved):
nice one :)

PR #828, opened 2026-04-30
feat(aiservices): add AI API Key management

loicrouillermonay:
This PR adds CLI commands for managing AI API keys under `exo ai-api-key`:
- `create --name <name> --scope <public|deployment-uuid>` - creates a new API key and prints the plaintext value
- `list` - displays a table with id, name, scope, and created-at
- `get <id>` - shows detailed view (without the value field)
- `update <id> [--name <name>] [--scope <scope>]` - updates an existing API key
- `delete <id>` - deletes an API key
- `rotate <id>` - rotates an API key and prints the new plaintext value
- `reveal <id>` - reveals the plaintext value of an existing API key

This PR is marked as draft, waiting for a dependency merge in [egoscale](https://github.com/exoscale/egoscale).

### DIFF
--- a/cmd/aiservices/aiservices.go
+++ b/cmd/aiservices/aiservices.go
@@ -2,6 +2,7 @@ package aiservices
 
 import (
 	exocmd "github.com/exoscale/cli/cmd"
+	"github.com/exoscale/cli/cmd/aiservices/apikey"
 	"github.com/exoscale/cli/cmd/aiservices/deployment"
 	"github.com/exoscale/cli/cmd/aiservices/model"
 	"github.com/spf13/cobra"
@@ -18,4 +19,5 @@ func init() {
 	// Attach subcommand groups
 	DedicatedInferenceCmd.AddCommand(model.Cmd)
 	DedicatedInferenceCmd.AddCommand(deployment.Cmd)
+	DedicatedInferenceCmd.AddCommand(apikey.Cmd)
 }
--- a/cmd/aiservices/apikey/apikey.go
+++ b/cmd/aiservices/apikey/apikey.go
@@ -1,0 +1,11 @@
+package apikey
+
+import (
+	"github.com/spf13/cobra"
+)
+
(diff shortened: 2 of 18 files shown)


arnaudgeiser:
Wouldn't it make sense to actually create `ai` subcommands? That would match the experience we have on the portal.
But I do think that `dedicated-inference` should be moved there as-well.

loicrouillermonay:
> Wouldn't it make sense to actually create `ai` subcommands? That would match the experience we have on the portal. But I do think that `dedicated-inference` should be moved there as-well.

cc @anamitev 

Right now it's `exo dedicated-inference ...` but `exo ai ...` is an alias working the same way (that I'm already always using because it's shorter).

For the next XIP about "Managed Inference GA" (not MVP) we could consider making `exo ai` the main one and `dedicated-inference` an alias of it because it will be a weird UX to create a Managed Inference key via `exo dedicated-inference ai-api-key create`

retrack:
Yes that would respect the umbrella concept as presented to customers and
consistent with dbaas for example.

In the absence of Ana, ok from us.


--
Antoine

habfast:
but that's already the case: 
```
$ exo ai --help
Dedicated AI inference management

Usage:
  exo dedicated-inference [command]

Aliases:
  dedicated-inference, ai, inference

Available Commands:
  deployment  Manage AI deployments
  model       Manage AI models

Flags:
  -h, --help   help for dedicated-inference
```

PS: Allowing myself to comment since I'm flagged as reviewer, and it showed up in my github notifications 
Screenshot 2026-09-18 at 15 08 30

arnaudgeiser:
> but that's already the case:

Yes, exactly, it's an old comment, but we'll have:

- `exo ai deployment list`
- `exo ai model list`
- `exo ai api-key list` 

So we are fine, but I would deprecate `dedicated-inference` as a top-level item (or at least make sure we do not advertise it).

PR #831, opened 2026-05-05
Migrate most list APIs to failure-resilient streaming utilities

habfast:
# Description

Follow-up to https://github.com/exoscale/cli/pull/826, finishes [sc-177228](https://app.shortcut.com/exoscale/story/177228/improve-zone-failure-handling)

Migrated:
- cmd/aiservices/model/model_list.go                                                     
- cmd/aiservices/deployment/instance_type_list.go                                                     
- cmd/compute/elastic_ip/elastic_ip_list.go
- cmd/compute/instance_pool/instance_pool_list.go
- cmd/compute/private_network/private_network_list.go
- cmd/compute/load_balancer/nlb_list.go
- cmd/compute/deploy_target/deploy_target_list.go
- cmd/compute/sks/sks_list.go
- cmd/compute/sks/sks_nodepool_list.go
- cmd/dbaas/dbaas_list.go

Skipped: 
- cmd/compute/instance_type/instance_type_list.go (would break sorting)
- cmd/compute/instance/instance_list.go (custom ToTable+sync.Map, don't want to get into it...)

Note that the Changelog is already updated from the previous PR! It doesn't need to be touched again

## Checklist

* [ ] ~~Changelog updated~~
* [x] Testing

## Testing

Manually ran every single command that has been updated against the exoscale-ops organization. Works smoothly. 
Screenshot 2026-05-05 at 16 38 14

### DIFF
--- a/cmd/aiservices/deployment/instance_type_list.go
+++ b/cmd/aiservices/deployment/instance_type_list.go
@@ -58,7 +58,11 @@ func runInstanceTypeList(c *InstanceTypeListCmd, stdout, stderr io.Writer) error
 	defer sink.Flush()
 
 	streamer := output.NewStreamer(InstanceTypeListItemOutput{}, stdout)
-	defer func() { _ = streamer.Close() }()
+	defer func() {
+		if err := streamer.Close(); err != nil {
+			_, _ = fmt.Fprintf(stderr, "error: %s\n", err)
+		}
+	}()
 
 	failed := utils.ForEveryZoneAsync(ctx, zones, globalstate.RequestTimeout, sink, true,
 		func(ctx context.Context, zone v3.Zone) error {
--- a/cmd/aiservices/deployment/instance_type_list.go
+++ b/cmd/aiservices/deployment/instance_type_list.go
@@ -1,8 +1,10 @@
 package deployment
 
 import (
+	"context"
 	"fmt"
-	"sort"
+	"io"
+	"os"
 	"strings"
 
 	exocmd "github.com/exoscale/cli/cmd"
@@ -14,16 +16,10 @@ import (
 )
 
 type InstanceTypeListItemOutput struct {
-	Family     string `json:"family"`
-	Authorized bool   `json:"authorized"`
-	Zone       string `json:"zone"`
+	Family     string `json:"family" outputWidth:"16"`
+	Authorized bool   `json:"authorized" outputWidth:"10"`
+	Zone       string `json:"zone" outputWidth:"8"`
 }
-
-type InstanceTypeListOutput []InstanceTypeListItemOutput
-
-func (o *InstanceTypeListOutput) ToJSON()  { output.JSON(o) }
-func (o *InstanceTypeListOutput) ToText()  { output.Text(o) }
-func (o *InstanceTypeListOutput) ToTable() { output.Table(o) }
 
 type InstanceTypeListCmd struct {
 	exocmd.CliCommandSettings `cli-cmd:"-"`
@@ -39,12 +35,17 @@ func (c *InstanceTypeListCmd) CmdLong() string {
 	return fmt.Sprintf(`This command lists AI instance types.
 
 Supported output template annotations: %s`,
-		strings.Join(output.TemplateAnnotations(&InstanceTypeListOutput{}), ", "))
+		strings.Join(output.TemplateAnnotations(&InstanceTypeListItemOutput{}), ", "))
 }
 func (c *InstanceTypeListCmd) CmdPreRun(cmd *cobra.Command, args []string) error {
 	return exocmd.CliCommandDefaultPreRun(c, cmd, args)
 }
+
 func (c *InstanceTypeListCmd) CmdRun(_ *cobra.Command, _ []string) error {
+	return runInstanceTypeList(c, os.Stdout, os.Stderr)
+}
+
+func runInstanceTypeList(c *InstanceTypeListCmd, stdout, stderr io.Writer) error {
 	ctx := exocmd.GContext
 	client := globalstate.EgoscaleV3Client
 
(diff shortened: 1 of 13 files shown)


pierre-emmanuelJ (approved):

cmd/aiservices/deployment/instance_type_list.go
```diff
@@ -53,45 +54,39 @@ func (c *InstanceTypeListCmd) CmdRun(_ *cobra.Command, _ []string) error {
...
-		return nil
-	})
+	streamer := output.NewStreamer(InstanceTypeListItemOutput{}, stdout)
+	defer func() { _ = streamer.Close() }()
 
-	sortInstanceTypeListOutput(out)
-
-	return c.OutputFunc(&out, err)
-}
+	failed := utils.ForEveryZoneAsync(ctx, zones, globalstate.RequestTimeout, sink, true,
```
natalie-o-perret, line 67: The old code sorted by zone then family before printing. 

With async fan-out the output order is now random across runs. 

Was that intentional? 

If so worth mentioning in the PR description so users know to expect it.

cmd/aiservices/deployment/instance_type_list.go
```diff
@@ -53,45 +54,39 @@ func (c *InstanceTypeListCmd) CmdRun(_ *cobra.Command, _ []string) error {
...
-				Zone:       string(zone.Name),
-			})
-		}
+	sink := utils.NewWarningSinkTo(stderr)
+	defer sink.Flush()
 
-		return nil
-	})
+	streamer := output.NewStreamer(InstanceTypeListItemOutput{}, stdout)
+	defer func() { _ = streamer.Close() }()
```
natalie-o-perret: If `streamer.Close()` returns an error (e.g. it fails to finalize the JSON array) we silently drop it and still return `nil`. 

The caller will think everything worked but the output may be truncated. 

Could we at least log it to stderr here?

cmd/aiservices/deployment/instance_type_list.go
```diff
@@ -53,45 +54,39 @@ func (c *InstanceTypeListCmd) CmdRun(_ *cobra.Command, _ []string) error {
...
-	sort.Slice(out, func(i, j int) bool {
-		if out[i].Zone < out[j].Zone {
-			return true
-		}
-		if out[i].Zone > out[j].Zone {
-			return false
-		}
-		return out[i].Family < out[j].Family
-	})
+	if failed > 0 {
```
natalie-o-perret, line 90: Uber-nit: This is a behaviour change worth calling out: before, zone errors printed a warning but the command still exited 0. Now it exits non-0. 

It's (imho) the right call, but it could break scripts that rely on the old behavior. 

Should we mention it in the changelog?

cmd/aiservices/deployment/instance_type_list_test.go
```diff
@@ -61,49 +61,26 @@ func TestInstanceTypeList(t *testing.T) {
...
+		t.Fatalf("invalid json: %v\nstdout: %s", err, outBuf.String())
 	}
-	if out[2].Zone != "ch-gva-2" || out[2].Family != "gpu-a100" {
-		t.Errorf("expected ch-gva-2/gpu-a100 third, got %s/%s", out[2].Zone, out[2].Family)
+	if len(rows) != 2 {
+		t.Fatalf("expected 2 instance types, got %d", len(rows))
 	}
-	if out[3].Zone != "ch-gva-2" || out[3].Family != "gpu-h100" {
-		t.Errorf("expected ch-gva-2/gpu-h100 fourth, got %s/%s", out[3].Zone, out[3].Family)
+	for _, r := range rows {
```
natalie-o-perret, line 81: The mock returns 2 specific entries (`gpu-a5000`/unauthorized and `gpu-a100`/authorized) but the test only checks the zone name now. 

Since we're in a single-zone mock the order is deterministic, so we could still check the family names and authorized values like the old test did.

cmd/aiservices/deployment/instance_type_list_test.go
```diff
@@ -61,49 +61,26 @@ func TestInstanceTypeList(t *testing.T) {
...
-			t.Errorf("expected first family gpu-a100, got %s", (*o)[0].Family)
-		}
-		if (*o)[1].Family != "gpu-a5000" {
-			t.Errorf("expected second family gpu-a5000, got %s", (*o)[1].Family)
-		}
-		return nil
-	}
-	if err := cmd.CmdRun(nil, nil); err != nil {
+	var outBuf bytes.Buffer
+	if err := runInstanceTypeList(cmd, &outBuf, nil); err != nil {
```
natalie-o-perret: Passing `nil` as stderr here will cause a panic if any warning gets buffered (i.e. if a zone fails), because `WarningSink.Flush` calls `fmt.Fprintf` on the writer unconditionally. 

 Should pass `&errBuf` here like the other test files do.

cmd/aiservices/model/model_list_test.go
natalie-o-perret: `TestModelListOutput_ToJSON` used to test calling `.ToJSON()` explicitly. 

Now it does exactly the same thing as `TestModelList` (run the command in json mode, check `len == 2`).

Could either remove it or make it actually test something different (e.g. check the exact JSON shape or field values).

cmd/aiservices/deployment/instance_type_list.go
habfast: yeah that's intentional. I didn't touch it in `exo compute instance instance-type` because I didn't want to risk breaking someone's flow, but this is for my service, [the documentation about this command hasn't even been published yet](https://github.com/exoscale/community/pull/737). (Almost) no one knows this command exists. It's fine to forget about it.

natalie-o-perret: Alright, noted :bowing_woman:

natalie-o-perret (approved):
In the grand scheme of things, fine by me, just a few nit's here and there.

cmd/aiservices/deployment/instance_type_list.go
habfast: No your analysis is wrong: in the current behavior, the status code is 1 and we only print warnings, even if some zones succeeded:
```
$ exo ai deployment instance-type
error: 3 errors occurred:
	* ListAIInstanceTypes: http client do: Get "https://ppapi-at-vie-1.exoscale.com/v2/ai/instance-type": GET https://ppapi-at-vie-1.exoscale.com/v2/ai/instance-type giving up after 5 attempt(s)
	* ListAIInstanceTypes: http client do: Get "https://ppapi-ch-dk-2.exoscale.com/v2/ai/instance-type": GET https://ppapi-ch-dk-2.exoscale.com/v2/ai/instance-type giving up after 5 attempt(s)
	* ListAIInstanceTypes: http client do: Get "https://ppapi-de-fra-1.exoscale.com/v2/ai/instance-type": GET https://ppapi-de-fra-1.exoscale.com/v2/ai/instance-type giving up after 5 attempt(s)
$ echo $?
1
```

With the new behavior, same status code, but we print the part that succeeded as well as the error messages:
```
$ bin/exo ai deployment instance-type
┼──────────────────┼────────────┼──────────┼
│ FAMILY           │ AUTHORIZED │ ZONE     │
┼──────────────────┼────────────┼──────────┼
│ gpua30           │ true       │ ch-gva-2 │
┼──────────────────┼────────────┼──────────┼
warning: zone ch-dk-2: ListAIInstanceTypes: http client do: Get "https://ppapi-ch-dk-2.exoscale.com/v2/ai/instance-type": context deadline exceeded
warning: zone at-vie-1: ListAIInstanceTypes: http client do: Get "https://ppapi-at-vie-1.exoscale.com/v2/ai/instance-type": context deadline exceeded
warning: zone de-fra-1: ListAIInstanceTypes: http client do: Get "https://ppapi-de-fra-1.exoscale.com/v2/ai/instance-type": context deadline exceeded
error: 3 zone(s) failed
$ echo $?
1
```

PR #843, opened 2026-06-12
feat(sks): add `--name` option to kubeconfig command

nhedger:
> [!NOTE]
>  **Disclosure**: 

# Description

This PR adds a `-n|--name` option to `exo compute sks kubeconfig`, allowing users to set a custom cluster and context name in the generated kubeconfig.

## Why this matters

Currently, the cluster and context names are set to the cluster UUID, making kubeconfigs harder to distinguish when users manage multiple configurations.

### Without `--name`

```
exo compute sks kubeconfig demo admin --zone de-fra-1

apiVersion: v1
clusters:
  - cluster:
      certificate-authority-data: <redacted>
      server: https://7fcf7a3d-da67-46e8-af25-858d1eaef682.sks-de-fra-1.exo.io:443
    name: 7fcf7a3d-da67-46e8-af25-858d1eaef682
contexts:
  - context:
      cluster: 7fcf7a3d-da67-46e8-af25-858d1eaef682
      user: admin
      
    name: 7fcf7a3d-da67-46e8-af25-858d1eaef682
current-context: 7fcf7a3d-da67-46e8-af25-858d1eaef682
kind: Config
preferences: {}
users:
  - name: admin
    user:
      client-certificate-data: <redacted>
      client-key-data: <redacted>

```

With `--name`

```
exo compute sks kubeconfig demo admin --zone de-fra-1 --name demo

apiVersion: v1
clusters:
    - cluster:
        certificate-authority-data: <redacted>
        server: https://7fcf7a3d-da67-46e8-af25-858d1eaef682.sks-de-fra-1.exo.io:443
      name: "demo"
contexts:
    - context:
        cluster: "demo"
        user: admin
      name: "demo"
current-context: "demo"
kind: Config
preferences: {}
users:
    - name: admin
      user:
        client-certificate-data: <redacted>
        client-key-data: <redacted>

```

## Why you may NOT want to merge this

You may prefer to implement this API-side rather than having the CLI rewrite the kubeconfig locally. I do not think there is anything inherently wrong with the CLI-based approach, but if you want to offer the same capability in the portal, the API may be the better place for it.

That said, this PR could still be merged as-is, and the CLI-local rewriting could later be replaced by an API-backed implementation.

## Checklist
(For exoscale contributors)

* [x] Testing

## Testing

- Automated tests as part of the implementation
- Manual testing

### DIFF
--- a/cmd/compute/sks/sks_kubeconfig.go
+++ b/cmd/compute/sks/sks_kubeconfig.go
@@ -6,7 +6,7 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
 
 	exocmd "github.com/exoscale/cli/cmd"
 	"github.com/exoscale/cli/pkg/globalstate"
@@ -23,6 +23,7 @@ type sksKubeconfigCmd struct {
 
 	ExecCredential bool        `cli-short:"x" cli-usage:"output an ExecCredential object to use with a kubeconfig user.exec mode"`
 	Groups         []string    `cli-flag:"group" cli-short:"g" cli-usage:"client certificate group. Can be specified multiple times. Defaults to system:masters"`
+	Name           string      `cli-short:"n" cli-usage:"cluster name to use in the generated kubeconfig"`
 	TTL            int64       `cli-short:"t" cli-usage:"client certificate validity duration in seconds"`
 	Zone           v3.ZoneName `cli-short:"z" cli-usage:"SKS cluster zone"`
 }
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	golang.org/x/text v0.31.0
 	gopkg.in/h2non/gentleman.v2 v2.0.4
 	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v3 v3.0.1
```
nhedger, line 36: I added `yaml.v3` as a direct dependency so we can cleanly generate double-quoted strings. Since it was already included transitively, I figured this was acceptable.

cmd/compute/sks/sks_kubeconfig.go
```diff
@@ -111,6 +112,10 @@ func (c *sksKubeconfigCmd) CmdPreRun(cmd *cobra.Command, args []string) error {
 }
 
 func (c *sksKubeconfigCmd) CmdRun(_ *cobra.Command, _ []string) error {
+	if c.ExecCredential && c.Name != "" {
+		return fmt.Errorf("--name cannot be used with --exec-credential")
+	}
+
```
nhedger, line 118: Being a bit defensive here, could also just ignore `--name` when using `--exec-credential`. I leave that to your appreciation.

pierre-emmanuelJ: Thanks for your change :)

You can try how the API reacts with both fields, and we adapt then here but can make sense here.

pierre-emmanuelJ (approved):

PR #847, opened 2026-06-19
feat(store): add bucket lifecycle support

mping-exo:
# Description

- Bump s3 dependency to fix some issues with serde (eg previous versions wouldn't serialize a field with `false` value because they were missing this [old check](https://github.com/aws/aws-sdk-go-v2/blame/v1.2.0/service/s3/serializers.go?rgh-link-date=2026-06-19T12%3A55%3A19Z#L9891) vs [new check](https://github.com/aws/aws-sdk-go-v2/blame/v1.23.1/service/s3/serializers.go#L10231)
  - this caused alot of small changes, mostly some fields are now pointers and referencing code had to be updated
  - I reworked the PR into two commits, 1 with proper changes and another just for vendor changes"
- Add bucket lifecycle commands to storage
  - set
  - show
  - delete

## Checklist
(For exoscale contributors)

* [x] Testing

### get/show/delete config

```

$ go run main.go -Approd -z ch-dk-2 storage bucket lifecycle set sos://miguel1 ../lifecycle-tests/aws2.json

$ go run main.go -Approd -z ch-dk-2 storage bucket lifecycle show sos://miguel1

┼──────────────────┼─────────────────────────────────────────────────────────────────────────────────┼
│ BUCKET LIFECYCLE │                                                                                 │
┼──────────────────┼─────────────────────────────────────────────────────────────────────────────────┼
│ Bucket           │ miguel1                                                                         │
│ Rules            │                                                                                 │
│                  │   ID                                      Expire-Objects-On-Specific-Date       │
│                  │   Status                                  Enabled                               │
│                  │   Filter (And) object-size-greater-than                                  1024   │
│                  │   Filter (And) object-size-less-than                            2048999999999   │
│                  │   Expiration date                         2025-12-31 00:00:00 +0000 UTC         │
│                  │   ID                                      Noncurrent-Version-Expiration-1-Day   │
│                  │   Status                                  Enabled                               │
│                  │   Noncurrent expiration days                                                1   │
│                  │   ID                                      Expire-Old-Delete-Markers             │
│                  │   Status                                  Enabled                               │
│                  │   Expire delete marker                    false                                 │
│                  │                                                                                 │
┼──────────────────┼─────────────────────────────────────────────────────────────────────────────────┼

$ go run main.go -Approd -z ch-dk-2 storage bucket lifecycle delete sos://miguel1
```

### expiration header
```
┼───────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────┼
│    STORAGE    │                                                                                                               │
┼───────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────┼
│ Path          │ some-key                                                                                                      │
│ Bucket        │ miguel1                                                                                                       │
│ Last Modified │ 2025-04-18 15:21:50 UTC                                                                                       │
│ Size          │ 10 B                                                                                                          │
│ URL           │ https://sos-ch-dk-2.exo.io/miguel1/some-key                                                                   │
│ ACL           │                                                                                                               │
│               │   Read           -                                                                                            │
│               │   Write          -                                                                                            │
│               │   Read ACP       -                                                                                            │
│               │   Write ACP      -                                                                                            │
│               │   Full Control   49bda34b-82ac-4686-a5f9-7b542cbfc649                                                         │
│               │                                                                                                               │
│ Metadata      │                                                                                                               │
│ Headers       │                                                                                                               │
│               │   Content-Encoding                                                                                            │
│               │   Content-Type       application/octet-stream                                                                 │
│               │   Expires-At         expiry-date="Wed, 31 Dec 2025 00:00:00 GMT", rule-id="Expire-Objects-On-Specific-Date"   │
│               │                                                                                                               │
┼───────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────┼
```


natalie-o-perret:
Thanks for your contribution, in grand scheme of things lgtm, I have one tiny issue with some enum handling in the default case, will leave a comment in the code directly.

I'm also wondering if the dep bump could have been separated from the feature, separate PRs.

pkg/storage/sos/lifecycle.go
```diff
@@ -0,0 +1,237 @@
...
+			},
+		}
+	case r.Filter.Prefix != nil:
+		return &types.LifecycleRuleFilterMemberPrefix{Value: *r.Filter.Prefix}
+	case r.Filter.ObjectSizeGreaterThan != nil:
+		return &types.LifecycleRuleFilterMemberObjectSizeGreaterThan{Value: *r.Filter.ObjectSizeGreaterThan}
+	case r.Filter.ObjectSizeLessThan != nil:
+		return &types.LifecycleRuleFilterMemberObjectSizeLessThan{Value: *r.Filter.ObjectSizeLessThan}
+	default:
+		return &types.LifecycleRuleFilterMemberPrefix{}
```
natalie-o-perret, line 94: Here we return `&types.LifecycleRuleFilterMemberPrefix{}` when none of the recognised fields are set. 

This silently turns any rule with an unrecognised filter into an empty prefix filter rather than erroring.

We might want to add some signal, wdyt?

tests/integ/with-api/lifecycle_test.go
natalie-o-perret, line 1: We test the happy path, but we might also want ToS3 => FromS3 round-trip test to catch silent field drops, `default:` case and any future divergence between the two.

mping-exo: @natalie-o-perret I will add more tests; can you show me where to view this test's results within the CI? Is it wired to run in CI?

pkg/storage/sos/lifecycle.go
mping-exo: The filter only has those 4 options (`and`, `prefix`, `sizegt`, `sizelt`); however an empty filter is allowed (will be a filter with an empty prefix - exactly whats required), but the `filter` should be present on all rules.
With that in mind I think this is OK

natalie-o-perret: It is, the empty filter => empty prefix => match-all semantics checks out against the current API, I'm just wondering if we could pin down a round-trip test (following up on the other thread) so any future SDK-side filter variant doesn't silently land here. Not blocking per se.

tests/integ/with-api/lifecycle_test.go
natalie-o-perret: Sorry to only get back you now, yes :bowing_woman:, even a small ToS3 => FromS3 round-trip covering today's 4 filter shapes (and the 0 rule case) would be enough, it'd catch silent field drops in today's shapes and pin the empty-filter behaviour from the other thread.

Again, not blocking, per se.

Running locally: no env vars needed if you have a default account configured.

Note `.github/workflows/e2e.yml` only covers compute + dbaas today, no storage entry.

Run:
```
cd tests/integ/with-api
go test -v -tags=integration_api -run TestLifecycle
```

The framework shells out to bin/exo, which reads `~/.config/exoscale/exoscale.toml` natively (documented in `tests/integ/README.md`).

Env vars can work as a fallback.

natalie-o-perret (approved):
The nit test can be done later in a follow-up pr, not blocking.

tests/integ/with-api/lifecycle_test.go
mping-exo: I added more tests in [2fb2f45](https://github.com/exoscale/cli/pull/847/commits/2fb2f45a97bff7b07bdcb15e0b940d3651ca9e6a)

natalie-o-perret: > I added more tests in [2fb2f45](https://github.com/exoscale/cli/pull/847/commits/2fb2f45a97bff7b07bdcb15e0b940d3651ca9e6a)

Perfect, ta!

pierre-emmanuelJ (approved):
thanks

PR #855, opened 2026-07-01
storage(lifecycle): use RegisterCLICommand pattern

natalie-o-perret:
# Description

Rewrites the lifecycle commands to use the struct-based `RegisterCLICommand` pattern, consistent with newer commands like `cmd/aiservices/deployment/`.

Changes:
- `set`, `show`, `delete` converted to typed structs with `RegisterCLICommand`
- Parent command simplified (redundant zone flag and helper func removed)

## Checklist
(For exoscale contributors)

* [ ] Changelog updated (under *Unreleased* block, and add the Pull Request #number for each bit you add to the `CHANGELOG.md`)
* [x] Testing

## Testing

Build and full test suite pass (`make build && make test-verbose`). Lint clean.


natalie-o-perret:
[SC-187336]

quentinalbertone (approved):
good idea ! Do you think that you can move the `storage_bucket_lifecycle*` into a subpackage ? You can find an example in the `aiservices` package

natalie-o-perret:
> good idea ! Do you think that you can move the `storage_bucket_lifecycle*` into a subpackage ? You can find an example in the `aiservices` package

I've just signed off, 🚇-typing, but the idea is that we can drive the people using "AI" for external contributions with the AGENTS.md, (https://github.com/exoscale/cli/pull/856)  file sitting now at the root of the repo.

Ironically-speaking, I precisely gave the AI services as an example to develop rule for the LLM to grasp the structure / pattern, and it didn't pick it on its own 😂, I will further adjust the rules there so that next time it picks it up.

Good catch anyway, thanks!

---

Addressed

pierre-emmanuelJ (approved):

PR #871, opened 2026-07-09
limits: detailed GPU quotas

simisoft-exo:
image

* [x] Tested locally

AI disclosure: used sandboxed Hermes Agent with Qwen 3.6 27B on local GPU

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 # Changelog
 
 ## Unreleased
-- limits: Added GPU by type to `exo limits` and subcommand for GPU only details with `exo limits gpu`
 
 ### Improvements
 - Upgrade aws sdk #874
 - Improve github action workflow #874
+- limits: Added GPU by type to `exo limits` and subcommand for GPU only details with `exo limits gpu` #871
 
 ## 1.95.6
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- limits: GPU quota details with `exo limits --gpu`
 
 ### Improvements
 - Upgrade aws sdk #874
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Unreleased
-- limits: GPU quota details with `exo limits --gpu`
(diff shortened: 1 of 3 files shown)


cmd/limits.go
```diff
@@ -42,6 +50,8 @@ func (o *LimitsOutput) ToJSON()  { output.JSON(o) }
 func (o *LimitsOutput) ToText()  { output.Text(o) }
 func (o *LimitsOutput) ToTable() { output.Table(o) }
 
+var showDetailedGPU bool
+
 var limitsCmd = &cobra.Command{
```
quentinalbertone, line 53: **nitpick (if-minor):** Can you follow the new way to declare command, you can find an example in the `cmd/aiservices` package

quentinalbertone (changes requested):
You need to address the warning given by `govulncheck`

cmd/limits.go
natalie-o-perret: > AI disclosure: used sandboxed Hermes Agent with Qwen 3.6 27B on local GPU

@simisoft-exo if you're using an agent, there is an [AGENTS.md](https://github.com/exoscale/cli/blob/master/AGENTS.md) that can usually help them, interesting that hermes didn't pick it up especially given their documentation https://hermes-agent.nousresearch.com/docs/user-guide/features/context-files#agentsmd :thinking:

simisoft-exo: Let's say I had other suggestions that were contrary to AGENTS.md and cli conventions.

simisoft-exo:
> You need to address the warning given by `govulncheck`

Addressed. Now other things don't work and I'm not familiar with the dependency tree to fix it.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- limits: Added GPU by type to `exo limits` and subcommand for GPU only details with `exo limits gpu`
```
quentinalbertone: **nitpick (blocking):** you need to add a line in the changelog under a subtitle like features, improvements, fixes, ...

quentinalbertone (changes requested):

CHANGELOG.md
simisoft-exo: fixed in: https://github.com/exoscale/cli/pull/871/changes/ab02aed304d8b85de4cc59ef1aeb0b9447858c73

quentinalbertone (approved):

PR #880, opened 2026-07-23
Vpc integration

aureliar8:
# Description

Heavily AI-assisted  but manually used & tested 

[ch190640]

The E2E tests are currently failing due to a bug in the orchestrator, we are fixing it at the source and I'll ensure those are green before merging 

## Checklist
(For exoscale contributors)

* [x] Testing


cmd/compute/instance/instance_show.go
```diff
@@ -124,20 +127,31 @@ func (c *instanceShowCmd) CmdRun(cmd *cobra.Command, _ []string) error {
 		PublicIPAssignment: instance.PublicIPAssignment,
 		IPAddress:          utils.DefaultIP(&instance.PublicIP, "-"),
 		IPv6Address:        utils.DefaultIP(ipV6, "-"),
+		//TODO: we need to fix the orchestrator to prevent NPEs here
+		Vpc: instance.Vpc.Name,
```
natalie-o-perret, line 132: Could we handle `instance.Vpc == nil` here?

The field is [optional](https://github.com/exoscale/cli/blob/c1325927a933d9cafec640b064b507056ae7c645/vendor/github.com/exoscale/egoscale/v3/schemas.go#L2961-L3022), and the api omits it for instances without a vpc. The [failing compute e2e](https://github.com/exoscale/cli/actions/runs/33627153655/job/100295571464) creates both instances, then panics here with exit 2. As `instance create` runs show after creation, this can report failure after the instance already exists.

The existing ci repro is:

```sh
cd tests/e2e
go test -v -tags=api -timeout 30m -run TestScriptsAPICompute 2>&1
```

```text
panic: runtime error: invalid memory address or nil pointer dereference
/home/runner/work/cli/cli/cmd/compute/instance/instance_show.go:134
[exit status 2]
```

cmd/networking/vpc/vpc_subnet_create.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return err
+	}
+
+	req := v3.CreateSubnetRequest{
+		Name:          c.Name,
+		Description:   c.Description,
+		Ipv4Block:     c.IPv4Block,
+		AddressSpace:  v3.CreateSubnetRequestAddressSpace(c.AddressSpace),
+		Addressfamily: v3.CreateSubnetRequestAddressfamily(c.AddressFamily),
```
natalie-o-perret, line 71: Could we default these to `inet4` and `private`, since they are the only values the client accepts?

Both flags currently default to empty strings, and the request always sends them. The [generated request](https://github.com/exoscale/cli/blob/c1325927a933d9cafec640b064b507056ae7c645/vendor/github.com/exoscale/egoscale/v3/operations.go#L20112-L20135) marks both fields required. A normal create with `--ipv4-block 10.20.0.0/24` reaches the api with `"address-space":""` and `"addressfamily":""`.

cmd/networking/vpc/vpc_route_create.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return err
+	}
+
+	subnetEntry, err := FindSubnet(ctx, client, vpcEntry.ID, c.Subnet)
+	if err != nil {
+		return err
+	}
+
+	if _, err := client.CreateRoute(ctx, vpcEntry.ID, subnetEntry.ID, v3.CreateRouteRequest{
```
natalie-o-perret, line 62: Could we render the route returned by `CreateRoute` instead of listing again?

The [api call](https://github.com/exoscale/cli/blob/c1325927a933d9cafec640b064b507056ae7c645/vendor/github.com/exoscale/egoscale/v3/operations.go#L20513-L20570) already returns the new route. If the follow-up lookup fails, the command exits non-zero after the route was created. On success it also prints inherited and existing routes, rather than only the new route id.

cmd/compute/instance/instance_show.go
```diff
@@ -18,29 +18,32 @@ import (
...
+	InstanceType       string                `json:"instance_type"`
+	Template           string                `json:"template"`
+	Zone               v3.ZoneName           `json:"zone"`
+	AntiAffinityGroups []string              `json:"anti_affinity_groups" outputLabel:"Anti-Affinity Groups"`
+	DeployTarget       string                `json:"deploy_target"`
+	SecurityGroups     []string              `json:"security_groups"`
+	PrivateInstance    string                `json:"private-instance" outputLabel:"Private Instance"`
+	ElasticIPs         []string              `json:"elastic_ips" outputLabel:"Elastic IPs"`
+	PublicIPAssignment v3.PublicIPAssignment `json:"public-ip" outputLabel:"Public IP"`
+	Vpc                string                `json:"vpc"`
```
natalie-o-perret, line 36: Would it make sense to append these fields to `InstanceShowOutput`?

[Text output follows declaration order](https://github.com/exoscale/cli/blob/c1325927a933d9cafec640b064b507056ae7c645/pkg/output/outputter.go#L47-L60), and the docs [suggest parsing it with `cut`](https://github.com/exoscale/cli/blob/c1325927a933d9cafec640b064b507056ae7c645/cmd/cmd.go#L641-L643). [v1.99.0](https://github.com/exoscale/cli/blob/916feced2bf0312bfd78a053d1426cf95adc47ca/cmd/compute/instance/instance_show.go#L20-L44) has `PrivateNetworks`, `ElasticIPs`, `PublicIPAssignment`, then `IPAddress`. Inserting fields here shifts every existing text column from `PrivateNetworks` onwards.

cmd/networking/vpc/vpc_list.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/exoscale/cli/pkg/output"
+	"github.com/exoscale/cli/utils"
+	v3 "github.com/exoscale/egoscale/v3"
+)
+
+type vpcListItemOutput struct {
+	ID          v3.UUID     `json:"id" outputWidth:"36"`
+	Name        string      `json:"name" outputWidth:"30"`
+	Zone        v3.ZoneName `json:"zone" outputWidth:"8"`
+	Default     string
```
natalie-o-perret: Could we keep `Default` as a bool and add `json:"default"` here?

List currently emits `"Default":"no"`, while [show emits](https://github.com/exoscale/cli/blob/c1325927a933d9cafec640b064b507056ae7c645/cmd/networking/vpc/vpc_show.go#L24-L32) `"default":"no"`. Also, `"no"` is non-empty, so `{{if .Default}}` is true for both default and non-default vpcs.

cmd/networking/vpc/vpc_show.go
```diff
@@ -0,0 +1,163 @@
...
+
+	t.Append([]string{"Zone", string(o.Zone)})
+	t.Append([]string{"Created At", o.CreatedAt})
+	t.Append([]string{"Labels", func() string {
+		if len(o.Labels) == 0 {
+			return "n/a"
+		}
+
+		pairs := make([]string, 0, len(o.Labels))
+		for _, k := range o.Labels {
```
natalie-o-perret: Could this be `for k, v := range o.Labels`?

The current loop uses each value as a key, so `{"env":"prod"}` renders as `prod:`. The [same loop is in subnet show](https://github.com/exoscale/cli/blob/c1325927a933d9cafec640b064b507056ae7c645/cmd/networking/vpc/vpc_subnet_show.go#L52-L61), so labels are wrong in both table views.

cmd/networking/vpc/vpc_subnet_create.go
aureliar8: I don't really want that because in 3 month we'll accept `inet6` and in 1 year we'll accept `"public"`

And when that happens there's no reason for `inet4`/`private` to be a default

cmd/compute/instance/instance_show.go
aureliar8: I've done it on purpose because I want the fields related to the connectivity of an instance to be bundled together (publicIP, publicIPv6, VPC, VPC subnets & Private networks)

cmd/networking/vpc/vpc_route_create.go
aureliar8: I think it makes sense to show all routing rules after an add. It's also what we do for eg security group where we show the parent sg and all its rules after a add/delete. Wdyt ?

cmd/networking/vpc/vpc_list.go
aureliar8: I though that seeing default: `yes/no` made more sense than `true/false`

```
┼──────────────────────────────────────┼────────────────────────────────┼──────────┼──────────────┼──────────────────────────────────────────┼
│ ID                                   │ NAME                           │ ZONE     │ DEFAULT      │ DESCRIPTION                              │
┼──────────────────────────────────────┼────────────────────────────────┼──────────┼──────────────┼──────────────────────────────────────────┼
│ 53df4f25-365e-4d48-be28-45f5de1c0bd1 │ my-first-vpc                   │ ch-gva-2 │ no           │                                          │
┼──────────────────────────────────────┼────────────────────────────────┼──────────┼──────────────┼──────────────────────────────────────────┼
```

vs 
```
┼──────────────────────────────────────┼────────────────────────────────┼──────────┼──────────────┼──────────────────────────────────────────┼
│ ID                                   │ NAME                           │ ZONE     │ DEFAULT      │ DESCRIPTION                              │
┼──────────────────────────────────────┼────────────────────────────────┼──────────┼──────────────┼──────────────────────────────────────────┼
│ 53df4f25-365e-4d48-be28-45f5de1c0bd1 │ my-first-vpc                   │ ch-gva-2 │ false        │                                          │
┼──────────────────────────────────────┼────────────────────────────────┼──────────┼──────────────┼──────────────────────────────────────────┼
```

But I can change it

cmd/compute/instance/instance_show.go
aureliar8: Actually we'll fix it at the API level so that the field is always provided & we don't need to do a nil check

cmd/networking/vpc/vpc_subnet_create.go
kobajagi: It's a valid argument by @natalie-o-perret , please address it. Both fields are required in OpenAPI spec and need to be marked as such in CLI or have a default value.

aureliar8: > Both fields are required in OpenAPI spec and need to be marked as such in CLI 

Ah ok maybe it's that part that I missed, so how do we mark CLI fields as required ?

cmd/networking/networking.go
```diff
@@ -0,0 +1,19 @@
+// AI-generated by claude-opus-5 - not reviewed yet
```
sauterp: Now it's reviewed and can be removed IMO :)
```suggestion
```

cmd/networking/vpc/vpc_route_create.go
```diff
@@ -0,0 +1,88 @@
...
+	exocmd.CliCommandSettings `cli-cmd:"-"`
+
+	_ bool `cli-cmd:"create"`
+
+	VPC string `cli-arg:"#" cli-usage:"VPC-NAME|ID"`
+
+	Subnet      string `cli-usage:"Subnet to create the route in (NAME|ID)"`
+	Destination string `cli-usage:"route destination CIDR (e.g. 10.9.0.0/24)"`
+	// TODO: Add a proper link to the doc here to explain what the target can be
+	Target      string      `cli-usage:"route target, as ip=<IP address> (e.g. ip=10.0.0.5)"`
```
sauterp, line 21: I think we could document this in the API spec https://openapi-v2.exoscale.com/operation/operation-create-route#operation-create-route-body-application-json-target
currently it's empty.

cmd/networking/networking.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	exocmd "github.com/exoscale/cli/cmd"
+	"github.com/spf13/cobra"
+)
+
+// NetworkingCmd is the root command for networking services.
+var NetworkingCmd = &cobra.Command{
+	Use:        "networking",
+	Short:      "Networking services management",
```
sauterp, line 11: Should we add a note in the descriptions of this command(and perhaps all VPC commands), indicating that these features are still BETA?

sauterp (approved):
I've been using these changes extensively during the past weeks. Consider it tested.

cmd/networking/vpc/vpc_subnet_create.go
kobajagi: There is no helper, document it as required in attribute description and fail early if flags are not set.

cmd/networking/vpc/vpc_list.go
natalie-o-perret: Thanks, it lgtm now :bowing-woman

cmd/networking/vpc/vpc_route_create.go
natalie-o-perret: Alright, makes sense.

cmd/networking/networking.go
nerzhul: i think it's a good idea

cmd/compute/instance/instance_show.go
natalie-o-perret: Alright, yea, that sounds better handled at the api level.

cmd/networking/vpc/vpc_subnet_create.go
aureliar8: Does c1989e0e work for you ?

cmd/compute/instance/instance_show.go
natalie-o-perret: Gotcha, thanks for clarifying the intent.

cmd/networking/vpc/vpc_route_create.go
aureliar8: Yeah we also should point to community there, but we need a proper article as it'll be quite a complex parameter

cmd/networking/networking.go
aureliar8: I don't know about this. My understanding was that we don't market VPC to be beta at all. It's either not there yet or available & we commit on it

So I think it make sense to just merge that & enable all VPC operation on october 31st. Which means that between the next release of the CLI & the MVP data customer can have a CLI that appears to support VPC but the API rejects all call 

Wdyt ?

cmd/networking/vpc/vpc_subnet_create.go
natalie-o-perret: Yea, that works for me. 

It keeps both choices explicit and fails early when either flag is missing, thanks! :bowing_woman:

natalie-o-perret (approved):
✔️ Approving ahead of time (conflicts, and comments left by others still need to be resolved / addressed) 🚀

cmd/networking/networking.go
sauterp: sounds good to me. I think as long as we don't announce VPC officially, customers won't try these commands anyway.

cmd/compute/instance/instance_show.go
```diff
@@ -124,20 +128,31 @@ func (c *instanceShowCmd) CmdRun(cmd *cobra.Command, _ []string) error {
 		PublicIPAssignment: instance.PublicIPAssignment,
 		IPAddress:          utils.DefaultIP(&instance.PublicIP, "-"),
 		IPv6Address:        utils.DefaultIP(ipV6, "-"),
+		//TODO: we need to fix the orchestrator to prevent NPEs here
```
rutgerbeltman, line 131: Done?

PR #903, opened 2026-08-31
iam: include policy in role show output

natalie-o-perret:
# Description

Show an IAM role policy alongside the normal role details. The policy is rendered as a separate table with a `Policy:` heading. Keep `--policy` as the existing policy-only output for scripts.

JSON output includes the policy under `policy`.

## Checklist
(For exoscale contributors)

* [x] Testing

## Testing

Live before/after check using a temporary role with a denied service and three policy rules. The role was deleted after capturing the output.

Before, using the installed binary:

```console
$ exo iam role show cli-pr-903-policy-output-test
| ID          | <role-id>                            |
| Name        | cli-pr-903-policy-output-test        |
| Description | Temporary CLI PR 903 output test     |
| Editable    | true                                 |
| Labels      | purpose:cli-pr-903                   |
| Permissions | n/a                                  |
```

After, using the patched binary:

```console
$ ./bin/exo iam role show cli-pr-903-policy-output-test
| ID          | <role-id>                            |
| Name        | cli-pr-903-policy-output-test        |
| Description | Temporary CLI PR 903 output test     |
| Editable    | true                                 |
| Labels      | purpose:cli-pr-903                   |
| Permissions | n/a                                  |

Policy:
| SERVICE | TYPE (DEFAULT STRATEGY "ALLOW") | RULE ACTION |                     RULE EXPRESSION                     |
|---------|---------------------------------|-------------|---------------------------------------------------------|
| sos     | rules                           | allow       | operation in ['list-sos-buckets-usage', 'list-buckets'] |
|         |                                 | deny        | !(resources.bucket in ['my-bucket', 'my-other-bucket']) |
|         |                                 | deny        | operation in ['list-objects', 'get-object']             |
| cdn     | deny                            |             |                                                         |
```

A live read-only check against the existing `DBaaS` role confirms the final blank-line and heading layout.

JSON output contains the complete policy under `policy`. `--policy` continues to print only the policy table.

A focused fake API test covers the reported deny-by-default policy with `compute` and `networking` allowed. `go test ./cmd/iam` passes.

The full `make test-verbose` suite passes. `make build` succeeds. `golangci-lint run --timeout 4m` reports `0 issues`.

---

> [!NOTE]
> AI assistance: PR description, test scaffolding.

### DIFF
--- a/cmd/iam/iam_role_show.go
+++ b/cmd/iam/iam_role_show.go
@@ -22,11 +22,47 @@ type iamRoleShowOutput struct {
 	Editable    bool              `json:"editable"`
 	Labels      map[string]string `json:"labels"`
 	Permissions []string          `json:"permission"`
+	Policy      *iamPolicyOutput  `json:"policy" output:"-"`
 }
 
-func (o *iamRoleShowOutput) ToJSON()  { output.JSON(o) }
-func (o *iamRoleShowOutput) ToText()  { output.Text(o) }
-func (o *iamRoleShowOutput) ToTable() { output.Table(o) }
+func (o *iamRoleShowOutput) ToJSON() { output.JSON(o) }
+func (o *iamRoleShowOutput) ToText() { output.Text(o) }
+func (o *iamRoleShowOutput) ToTable() {
+	output.Table(o)
+	if o.Policy != nil {
+		o.Policy.ToTable()
+	}
+}
+
+func iamPolicyToOutput(policy *v3.IAMPolicy) *iamPolicyOutput {
+	if policy == nil {
+		return nil
+	}
+
+	out := &iamPolicyOutput{
+		DefaultServiceStrategy: string(policy.DefaultServiceStrategy),
+		Services:               map[string]iamPolicyServiceOutput{},
+	}
+
+	for name, service := range policy.Services {
+		rules := []iamPolicyServiceRuleOutput{}
+		if service.Type == "rules" {
+			for _, rule := range service.Rules {
+				rules = append(rules, iamPolicyServiceRuleOutput{
+					Action:     string(rule.Action),
+					Expression: rule.Expression,
+				})
+			}
+		}
+
+		out.Services[name] = iamPolicyServiceOutput{
+			Type:  string(service.Type),
+			Rules: rules,
+		}
+	}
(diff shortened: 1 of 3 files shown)


natalie-o-perret:
[SC-196302]

kobajagi:
Can you add example of the output table with full featured policy?

natalie-o-perret:
Updated the PR description accordingly with output from a live full-featured policy, including allow and deny rules with expressions.

natalie-o-perret:
> Can you add example of the output table with full featured policy?

Updated the PR description accordingly with output from a live full-featured policy, including allow and deny rules with expressions.

kobajagi:
Thanks @natalie-o-perret . IMO it feels cluttered, rule expression can also be very long which will extend policy table even more. Probably why policy was factored out in the first place.

cmd/iam/iam_role_show.go
```diff
@@ -111,6 +127,7 @@ func (c *iamRoleShowCmd) CmdRun(_ *cobra.Command, _ []string) error {
 		Labels:      role.Labels,
 		Name:        role.Name,
 		Permissions: role.Permissions,
+		Policy:      policy,
```
kobajagi, line 131: Can we instead print Policy as a fully separate table? With `\nPolicy:\n` in between.

kobajagi: It feels better UX for me, but feel free to disagree.

natalie-o-perret:
> IMO it feels cluttered, rule expression can also be very long which will extend policy table even more. Probably why policy was factored out in the first place.

Yea, agreed. I updated the PR to keep the default table unchanged and include the nested policy only in JSON output. `--policy` remains the opt-in full policy table.

cmd/iam/iam_role_show.go
natalie-o-perret: Sorry, I missed these inline comments when I replied to the conversation thread.

I've updated the output to render the policy as a separate table with a blank line and `Policy:` heading.

kobajagi: Thanks, i think it looks better now

kobajagi (approved):

